PR #2, opened 2019-08-02
Refactor/typescript migration

TheGame2500:


TheGame2500:
@edwardupton still gotta figure out how to move type definitions out of index.d.ts and into separate files as per convention. but I did catch a few bugs on this fun ride

PR #6, opened 2019-09-13
Add pageview event

edwardupton:


edwardupton:
[Shopify tracking script sends user ID, and GTM event for pageview](https://trello.com/c/NMUyZlNl/3202-shopify-tracking-script-sends-user-id-and-gtm-event-for-pageview)

PR #7, opened 2019-10-03
Add context to analytics.track

edwardupton:


edwardupton:
[[Segment] admin I can see which events come from Littledata](https://trello.com/c/sFktIzsW/3217-segment-admin-i-can-see-which-events-come-from-littledata)

PR #12, opened 2019-10-15
Always send actual clientId

TheGame2500:


TheGame2500:
[P1 ||[Bug] [Urgent] GA client ID is 'undefined' for server-side tracking events](https://trello.com/c/mW4lCLLJ/3328-bug-whenever-possible-ga-client-id-is-undefined-for-add-remove-from-cart-events)

PR #15, opened 2019-11-01
Fix/segment page callback

edwardupton:


edwardupton:
[[Bug] [Segment] Events/pageviews not firing](https://trello.com/c/h0mXnxhF/3191-bug-segment-events-pageviews-not-firing)

PR #16, opened 2019-11-12
Chunk product list views in 20

edwardupton:


edwardupton:
[P3 || On a Shopify product list page, only send maximum of 20 impressions in one hit to GA](https://trello.com/c/NyMhWWQU/3380-p3-on-a-shopify-product-list-page-only-send-maximum-of-20-impressions-in-one-hit-to-ga)

PR #17, opened 2019-11-12
Improvement/store client id metafield

TheGame2500:


TheGame2500:
[P2 || [Transaction-monitor] if I generate a 'default.littledata' clientID, I update customer attributes in shopify, so that shopify-tracker can initialise to generated clientID next time](https://trello.com/c/s857Hzhz/3375-p2-transaction-monitor-if-i-generate-a-defaultlittledata-clientid-i-update-customer-attributes-in-shopify-so-that-shopify-tracke)

PR #18, opened 2019-11-21
Fix/non ga segment anonmyous

edwardupton:


edwardupton:
[P1 || [Bug] Segment customers without GA destination not getting correct anonymous ID](https://trello.com/c/IgogiYtu/3400-p1-bug-segment-customers-without-ga-destination-not-getting-correct-anonymous-id)

PR #19, opened 2019-12-03
Fix product lists for Segment

edwardupton:


edwardupton:
[[Bug] Segment products list clicks have different list_id to product views](https://trello.com/c/EImWX09N/3424-bug-segment-products-list-clicks-have-different-listid-to-product-views)

PR #22, opened 2020-01-27
Improvement/segment anonymous

edwardupton:


edwardupton:
[P1 || Switch Segment app to use anonymous ID for client and server side](https://trello.com/c/DsDf2DQJ/3468-p1-switch-segment-app-to-use-anonymous-id-for-client-and-server-side)

PR #28, opened 2020-03-14
Change to Shopify (Littledata)

edwardupton:


edwardupton:
[Change Event Category in GA to use Littledata brand](https://trello.com/c/ssf0WCbR/3637-change-event-category-in-ga-to-use-littledata-brand)

PR #29, opened 2020-03-15
Improvement/pageview on url change

edwardupton:


edwardupton:
[[Shopify tracker] Trigger pageview and product view event for every change of URL](https://trello.com/c/AThwtYZz/3638-shopify-tracker-trigger-pageview-and-product-view-event-for-every-change-of-url)

PR #34, opened 2020-05-01
Improvement/remove set client id loop

edwardupton:


edwardupton:
[[Shopify Tracker] ClientId set just after add_to_cart event is being sent](https://trello.com/c/rsA5mVsM/3759-shopify-tracker-clientid-set-just-after-addtocart-event-is-being-sent)

PR #41, opened 2020-06-22
[R] fix plp tracking for some stores

MihaiCernat:


MihaiCernat:
[P1 || [Bug] [As soon as possible] Product List Impressions not Tracked for CUUP (PLP)](https://trello.com/c/734FJzb7/3867-p1-bug-as-soon-as-possible-product-list-impressions-not-tracked-for-cuup-plp)

PR #46, opened 2020-07-01
Filter GA product fields

edwardupton:


edwardupton:
[Add productId, compareAtPrice and variantId to Shopify dataLayer](https://trello.com/c/encmHhA3/3962-add-productid-compareatprice-and-variantid-to-shopify-datalayer)

src/gaTracker/helpers.ts
```diff
@@ -166,6 +167,26 @@ export const trackEvents = () => {
...
+		'category',
+		'variant',
+		'list',
+		'list_name',
+		'position',
+		'list_position',
+	];
+	const gaProduct = {} as LooseObject;
+	gaProductFields.forEach(field => {
+		if (gaProduct[field]) product[field] = gaProduct[field];
```
fstefan7, line 185: @edwardupton Won't `gaProduct` be always empty?

Secondly, this method does too many things, as it changes `product[field]` but also returns `gaProduct`. Also, it's hard to understand its purpose
I would suggest to change it to return one of the objects and to add another function if needed to update the other

edwardupton: Yes I got it wrong. I think it's ok to have one function now i've switched them around.

PR #49, opened 2020-07-06
bump version

MihaiCernat:


MihaiCernat:
[P2 || [Bug] Product List Impressions not Tracked for CUUP (PLP)](https://trello.com/c/734FJzb7/3867-p2-bug-product-list-impressions-not-tracked-for-cuup-plp)

PR #52, opened 2020-07-08
try to find clientId until ga is loaded

fstefan7:


fstefan7:
[P1 || [Bug] [As soon as possible] Client ID is lost on the server-side. Almost all transactions are from Clifton with (not set) Direct](https://trello.com/c/lTrRyVFu/3757-p1-bug-as-soon-as-possible-client-id-is-lost-on-the-server-side-almost-all-transactions-are-from-clifton-with-not-set-direct)

src/gaTracker/helpers.ts
```diff
@@ -23,6 +23,26 @@ export const initGtag = () => {
 	gtag('config', LittledataLayer.webPropertyID, getConfig());
 };
 
+let postClientIdTimeout: any;
+let nextTimeout = 500; // half a second
+const maximumTimeout = 524288000; // about 6 hours in seconds
+
+function waitForGaToLoad() {
+	const trackers = window.ga.getAll();
```
TheGame2500, line 31: @fstefan7 if ga is not loaded, this will throw and error. I suggest `const trackers = window.ga && window.ga.getAll && window.ga.getAll()` to ensure this never throws

fstefan7: Actually it doesn't. It returns an empty array.

These are the logs that I added:
```function waitForGtag() {
  var trackers = window.ga.getAll();
	console.log('trackers',trackers);
  if (trackers && trackers.length) {
    console.log('first');
    return setClientID(getGtagClientId, 'google');
  }
    console.log('second');
  if (nextTimeout > maximumTimeout) return;
  nextTimeout *= 2;
  clearTimeout(postClientIdTimeout);
  postClientIdTimeout = window.setTimeout(function () {
    console.log('new date', new Date());
    waitForGtag();
  }, nextTimeout);
}
```

This is the way they are printed:
```
trackers []
(index):673 second
(index):668 trackers (2) [pc, pc]
(index):670 first
```

src/gaTracker/helpers.ts
```diff
@@ -23,6 +23,26 @@ export const initGtag = () => {
 	gtag('config', LittledataLayer.webPropertyID, getConfig());
 };
 
+let postClientIdTimeout: any;
+let nextTimeout = 500; // half a second
+const maximumTimeout = 524288000; // about 6 hours in seconds
```
edwardupton, line 28: 6 hours?? Maybe 1 minute

edwardupton: I mean no one stays on a web page for 6 hours so it is academic

fstefan7: @edwardupton The reasoning was that even if the limit is so big (6 hours) we only try it around 28 times in 6 hours, so no we're not loading the website.
I could change it to whatever other value if you'll like to. Not sure though that 1 minute is enough.

edwardupton:
I'm ok with leaving it

PR #54, opened 2020-07-14
Validate before posting

edwardupton:


edwardupton:
[P1 || Invalid client IDs of true and 252525252 sent from Shopify tracker](https://trello.com/c/Mz85Ailx/3979-p1-invalid-client-ids-of-true-and-252525252-sent-from-shopify-tracker)

src/common/getCookie.ts
```diff
@@ -8,12 +8,13 @@ export const getCookie = (name: string): string => {
...
-			}
 			return cookie;
 		}
 	}
 	return '';
 };
+
+export const isGAclientIdValid = (cookie: string): string => {
+	const match = cookie.match(/(\d{2,11})\.(\d{2,11})/g);
+	return match && match[0];
```
fstefan7, line 19: @edwardupton This method should return a boolean, not a string.

edwardupton: Agreed

src/common/getCookie.ts
```diff
@@ -8,12 +8,13 @@ export const getCookie = (name: string): string => {
...
-				const match = cookie.match(/(\d{2,11})\.(\d{2,11})/g);
-				return match ? match[0] : '';
-			}
 			return cookie;
 		}
 	}
 	return '';
 };
+
+export const isGAclientIdValid = (cookie: string): string => {
```
fstefan7, line 17: @edwardupton and here the return type should be `bool`

src/common/getCookie.ts
```diff
@@ -8,12 +8,13 @@ export const getCookie = (name: string): string => {
 				cookieEnd = document.cookie.length;
 			}
 			const cookie = unescape(document.cookie.substring(valueStart, cookieEnd));
-			if (name === '_ga') {
```
TheGame2500, line 11: @edwardupton this will break carthook's usage of `getCookie`. Carthook expects this to return a clientID instead of `GA1.2.91802749.41897498`

src/segmentTracker/segmentTracker.ts
```diff
@@ -19,10 +19,7 @@ import { identifyCustomer, trackEvents, initSegment, callSegmentPage } from './h
 				window.ga(() => {
 					const tracker = window.ga.getAll()[0];
 					if (tracker) {
-						const clientId = tracker.get('clientId');
-						const generatedClientID =
-							LittledataLayer.customer && LittledataLayer.customer.generatedClientID;
-						const getClientID = () => (generatedClientID ? generatedClientID : clientId);
+						const getClientID = () => tracker.get('clientId');
```
TheGame2500, line 22: @edwardupton shouldn't this also check if client id is valid?

edwardupton: I thought about that, but I think not because it is later validated in postClientID(). i.e. we only use getClientID() in postClientID().

src/common/getCookie.ts
edwardupton: Had to switch function to `getValidGAClientId`

src/common/getCookie.ts
edwardupton: Yes thanks for spotting this. Renamed the function to `getValidGAClientId` for clarity

src/gaTracker/helpers.ts
```diff
@@ -217,5 +216,12 @@ export const getConfig = (): Gtag.CustomParams => {
 		user_id: userId,
 	};
 
+	const cookie = getCookie('_ga');
+	if (cookie && !isGAclientIdValid(cookie)) {
```
TheGame2500, line 220: @edwardupton this function does not seem to exist anymore after latest changes

edwardupton: Not sure why my linter didn't alert, but yes fixed

PR #60, opened 2020-07-24
added retrieval of clientID retry logic

fstefan7:


fstefan7:
[P1 || [Bug] [As soon as possible] Client ID is lost on the server-side. Almost all transactions are from Clifton with (not set) Direct](https://trello.com/c/lTrRyVFu/3757-p1-bug-as-soon-as-possible-client-id-is-lost-on-the-server-side-almost-all-transactions-are-from-clifton-with-not-set-direct)

PR #64, opened 2020-08-18
Fix/post fresh cart

edwardupton:


edwardupton:
[Fix cart storage since Shopify dropped cart token from our data layer](https://trello.com/c/Fm3VscrE/4096-fix-cart-storage-since-shopify-dropped-cart-token-from-our-data-layer)

src/common/helpers.ts
```diff
@@ -159,10 +162,8 @@ export function setClientID(getClientId: () => string, platform: 'google' | 'seg
 		const timePassed = Date.now() - Number(clientIdCreated);
 		// only need to resend cart if it's expired from our Redis cache
 		if (timePassed > timeout) {
-			postCartToLittledata(cart);
-			setTimeout(() => {
-				postClientID(getClientId, platform);
-			}, 10000); // allow 10 seconds for our server to register cart until updating it, otherwise there's a race condition between storing and a webhook triggered by this
```
TheGame2500, line 165: @edwardupton we're going to lose this 10000 delay, but i'm not sure what the race condition it prevents led to

edwardupton: Neither am I, but in the majority of cases postClientID would already have been called on line 151. The only situation it would _not_ have already been called is when `google-clientID` property does not exist but somehow `updatedAt` does.

I imagine adding cart attributes triggers a cart/update webhook after 10 seconds, but by which time the cart storage should already be registered in Redis - and so no extra add to cart will be sent.

Agreed?

TheGame2500 (approved):

PR #72, opened 2020-09-07
Handle MPEndpoint

edwardupton:


edwardupton:

src/gaTracker/helpers.ts
```diff
@@ -252,5 +243,11 @@ export const getConfig = (): Gtag.CustomParams => {
 		config.cookie_expires = 0;
 	}
 
+	if (LittledataLayer.MPEndpoint) {
```
fstefan7, line 246: @edwardupton Is there any possibility `LittledataLayer.MPEndpoint` to exist, but not to be an array? If so, then it will throw an error on line 249.
You can convert the if to 
```
if (LittledataLayer.MPEndpoint && LittledataLayer.MPEndpoint.length)
```
to be sure that we won't face this situation


Also, usually it's better to have the return statement as soon as possible, something like:
```
if (!LittledataLayer.MPEndpoint || !LittledataLayer.MPEndpoint.length) return config;

const length = LittledataLayer.MPEndpoint.length;
config.transport_url = LittledataLayer.MPEndpoint.slice(0, length - '/collect'.length);
return config;
```

This also prevents a lot of nested if-else branches

TheGame2500 (approved):

PR #75, opened 2020-09-10
fix substring

MihaiCernat:


MihaiCernat:
[P1 || [Bug] Pixel Perfect - InitiateCheckout sets wrong value](https://trello.com/c/NKUjUVIS/3994-p1-bug-pixel-perfect-initiatecheckout-sets-wrong-value)

ionescurares (approved):

PR #77, opened 2020-09-16
Extra referral exclusion

edwardupton:


edwardupton:
[Exclude shop.app as a referrer](https://trello.com/c/tT5RSz0t/4146-exclude-shopapp-as-a-referrer)

TheGame2500 (approved):

PR #79, opened 2020-09-18
Add cross-domain exclusion

edwardupton:


edwardupton:
[Exclude shop.app as a referrer and fix utm_source only referrals](https://trello.com/c/tT5RSz0t/4146-exclude-shopapp-as-a-referrer-and-fix-utmsource-only-referrals)

PR #91, opened 2020-10-20
Feature/measurement protocol transport url

edwardupton:


edwardupton:

src/gaTracker/helpers.ts
```diff
@@ -34,17 +28,24 @@ export const initGtag = () => {
...
+		...getConfig(),
+		send_page_view: false,
+	});
 };
 
 let postClientIdTimeout: any;
-let nextTimeout = 250;
-const maximumTimeout = 524288000; // about 6 hours in seconds
+let nextTimeout = 10;
+const maximumTimeout = 500;
```
fstefan7, line 43: @edwardupton I don't see the benefit of decreasing the value of `maximumTimeout`. As it is now, it means that if in half a second the GA library isn't loaded, then it won't be loaded anymore.
The previous implementation was to try to load the GA library for about 6 hours(there are under 30 calls made in that period because every time the nextTimeout period is doubling)

edwardupton:
The `waitForGAtoLoad` function is only triggered after GA is initialised
(within a ga callback). Waiting 250 ms (or even 6 hours) makes no sense, as
all we are waiting for is the `getAll` method to be available. Annoyingly
that is not when the callback is fired, but from testing around 10ms to
20ms later. Hence the setTimeout

src/gaTracker/helpers.ts
fstefan7: @edwardupton I understand why the change from 250ms to 10ms. The problem is that we won't know for sure if there will be an edge case for which the `getAll` method won't be available even after 500ms.(this is due to the fact that it is a client side method)

fstefan7: @edwardupton Anyway, from what I tested the setTimeout method added by me did not have a big impact for the clientID throughput, so I'm ok with changing the maximum period from 6h to 500ms. I think though that some throughput checking is needed(before and after the change), to know for sure that we're not affecting it

fstefan7 (approved):

PR #98, opened 2020-11-10
Fix/missing referrer

edwardupton:


edwardupton:
[Exclude self-referrals from Shopify tracking](https://trello.com/c/Ts1Xi5Vb/4288-exclude-self-referrals-from-shopify-tracking)

TheGame2500 (approved):

PR #101, opened 2020-11-12
Feature/email client

edwardupton:


edwardupton:
[Capture email from web forms and send with client and server Segment events](https://trello.com/c/7fPQCN88/4295-capture-email-from-web-forms-and-send-with-client-and-server-segment-events)

fstefan7 (approved):
Looks good, but I think the dist files shouldn't be included

PR #107, opened 2020-11-19
mapped ga3 client side events to ga4

fstefan7:


fstefan7:
[Client side parameter mapping between GA3 and GA4](https://trello.com/c/aM7DNc1g/4279-client-side-parameter-mapping-between-ga3-and-ga4)

src/gaTracker/helpers.ts
```diff
@@ -252,3 +205,106 @@ const addUTMMediumIfMissing = (url: string) => {
...
+			items: convertProductsToGa4Format(products),
+			item_list_name: listName,
+			item_list_id: listName,
+		});
+	} else {
+		const gaProducts = products.map(product => filterGAProductFields(product));
+		gtag('event', 'view_item_list', {
+			event_category,
+			items: gaProducts,
+			send_to: LittledataLayer.webPropertyID,
```
edwardupton, line 222: I think send_to is still needed with GA4: https://developers.google.com/analytics/devguides/collection/ga4/translate-events#if_you_currently_use_gtagjs

src/gaTracker/helpers.ts
```diff
@@ -252,3 +205,106 @@ const addUTMMediumIfMissing = (url: string) => {
 
 	return url.replace(sourceTag, utmTags);
 };
+
+function sendViewItemListEvent(products: Impression[]): void {
+	if (isGa4Format()) {
+		const listName = (products && products.length && products[0].list_name) || '';
+		gtag('event', 'view_item_list', {
```
edwardupton, line 212: In each case it is the event properties which are changing, not the name. So I would refactor to have a `getViewItemListProperties` function, which decides on the properties depending on `isGa4Format()`

src/gaTracker/helpers.ts
edwardupton: We should allow sending to BOTH GA3 and GA4. So if both webPropertyID and measurementID are in the data layer, send the same event to both.

src/gaTracker/helpers.ts
```diff
@@ -252,3 +205,106 @@ const addUTMMediumIfMissing = (url: string) => {
...
+			event_callback() {
+				window.clearTimeout(self.timeout);
+				document.location.href = self.href;
+			},
+		});
+	}
+}
+
+function isGa4Format(): boolean {
+	return LittledataLayer.webPropertyID && LittledataLayer.webPropertyID.includes('G-');
```
edwardupton, line 294: No, it should be called `LittledataLayer.measurementID` for consistency. Don't rely on parsing the string

edwardupton (changes requested):

edwardupton (approved):

edwardupton (approved):
convertProductsToGa4Format could really do with unit tests, but i'll approve this to add later

fstefan7:
@edwardupton I'll add a card for adding unit tests for the mapping

PR #113, opened 2020-12-15
use GA library when gtag throws an error for retrieving the clientId

fstefan7:


fstefan7:
[[GA4 & UA] ga/segment tracker to use gtag.js API](https://trello.com/c/1fqkaqrS/4285-ga4-ua-ga-segment-tracker-to-use-gtagjs-api)

src/common/helpers.ts
```diff
@@ -246,13 +246,23 @@ export function retrieveAndStoreClientId(withCustomTask: boolean = false) {
...
+				setCustomTask();
+			}
 
-		return setClientID(clientId, 'google');
-	});
+			return setClientID(clientId, 'google');
+		})
+		.catch(() => {
+			window.ga(() => {
+				const tracker = window.ga.getAll()[0];
```
edwardupton, line 259: I think this will run into a problem where window.ga is a function but getAll() returns nothing.

It needs to wait 10ms for library, see https://github.com/littledata/shopify-tracker/pull/91/files

edwardupton (changes requested):

edwardupton (approved):

PR #117, opened 2020-12-22
renamed measurementID field, increased version

fstefan7:


fstefan7:
[Client side parameter mapping between GA3 and GA4](https://trello.com/c/aM7DNc1g/4279-client-side-parameter-mapping-between-ga3-and-ga4)

ionescurares (approved):

PR #119, opened 2021-01-06
debug ga4 data sent

littlesorin:


littlesorin:
[Client side parameter mapping between UA and GA4](https://trello.com/c/aM7DNc1g/4279-client-side-parameter-mapping-between-ua-and-ga4)

fstefan7 (approved):

PR #144, opened 2021-02-14
PD-216 Track impressions without data layer

edwardupton:


src/common/productListViews.ts
```diff
@@ -1,63 +1,146 @@
...
+				});
+			})
+			.catch(ex => {
+				console.debug('Littledata unable to fetch', handle, ex);
+			}),
+	);
+};
+
+export const productAlreadyViewed = (handle: string, shopify_variant_id: string) =>
+	findProductInArray(LittledataLayer.ecommerce.impressions, handle, shopify_variant_id) ||
```
fstefan7, line 131: @edwardupton Won't this be true for all cases since arrays are `truthy`?

edwardupton: findProductInArray returns array.find(). If the product is not in either array then the result will be undefined. Could I make this clearer

fstefan7: Ah, then all good. My bad, I though that `findProductInArray` is returning array.filter() which is returning empty array if there is no match

fstefan7 (approved):

PR #176, opened 2021-03-12
PD-267 Send client id from thank you pages

TheGame2500:


src/gaThankYouPageTracker/gaThankYouPageTracker.ts
```diff
@@ -38,5 +39,22 @@ declare let window: CustomWindow;
 			value,
 			send_to: webPropertyId,
 		});
+
+		const shopId = getQueryStringParam(scriptSrc, 'shopId');
+		const env = getQueryStringParam(scriptSrc, 'env');
+		if (!shopId) return;
+
+		const transactionWatcherURLRoot =
+			env === 'production' ? 'https://transactions.littledata.io' : 'https://transactions-staging.littledata.io';
```
edwardupton, line 48: Define these in a constants file

src/segmentThankYouPageTracker/segmentThankYouPageTracker.ts
```diff
@@ -59,5 +60,21 @@ declare let window: ShopifyWindow;
...
+		const { user } = analytics;
+		const shopId = getQueryStringParam(scriptSrc, 'shopId');
+		const env = getQueryStringParam(scriptSrc, 'env');
+		if (!shopId || !user) return;
+
+		const transactionWatcherURLRoot =
+			env === 'production' ? 'https://transactions.littledata.io' : 'https://transactions-staging.littledata.io';
+		const clientIDEndpoint = `${transactionWatcherURLRoot}/v3/clientID/store/${shopId}`;
+
+		httpRequest.postJSON(clientIDEndpoint, {
```
edwardupton, line 74: Should we also send GA client ID for Segment? I know we can look it up on our server, but is it safer?

src/gaThankYouPageTracker/gaThankYouPageTracker.ts
edwardupton: And this is the wrong way around. The script should be backwards compatible with previous script tags, so if `env` is undefined it should send to production.

We can assume staging stores will always have up to date script tags.

edwardupton (changes requested):

src/segmentThankYouPageTracker/segmentThankYouPageTracker.ts
edwardupton: @TheGame2500 ?

edwardupton (approved):

PR #185, opened 2021-03-25
replace all occurances of email

MihaiCernat:


edwardupton (approved):
Good catch. You could also add a unit test for multiple emails

PR #201, opened 2021-04-16
PD-274 use json api

MihaiCernat:


edwardupton (changes requested):
Can you add a unit test, as I don't think you've got the output correct. `getJSON` returns `{ ip }` not `ip`.

See example in /test/common/getProductDetail.test.ts

edwardupton (approved):

PR #205, opened 2021-04-26
PD-79 Consent tracking

edwardupton:


src/common/helpers/getTrackingConsent.ts
```diff
@@ -0,0 +1,34 @@
+import { setClientID } from '../setClientID';
+import { CustomWindow } from '../../..';
+
+declare let window: CustomWindow;
+
+const clientIdIfNotTracked = 'userNotTracked';
+const performanceCookieCategory = '2,';
+
+export const getTrackingConsent = (platform: 'google' | 'segment'): Promise<void> => {
```
lyykfi, line 9: the problem is with the function is overcomplexity.
we should separate some home OptanonActiveGroups  and customerPrivacy login.

I would suggest:
`
const consents = [
 [() => typeof window.OptanonActiveGroups === 'string', ... function for OptanonActiveGroups ],
 [() => window.Shopify.customerPrivacy, ... function for customerPrivacy],
]

if (!consents.every((item) => {
 const [ validator, fnc] =  item;

if (validator) {
 fnc();
setClientID(clientIdIfNotTracked, platform);
}

})) {
resolve();
}
`

what do you think?

lyykfi (changes requested):

src/common/helpers/getTrackingConsent.ts
edwardupton: I've split into helper functions for each consent platform, but for just two platforms the if ... else seems ok

src/common/setClientID.ts
```diff
@@ -74,7 +74,7 @@ const checkCartHasAttributes = (cart: Cart.RootObject): Promise<void | Cart.Root
 	// until the attributes are added to cart, the cart token is not stable
 	const attributesToSet = Object.keys(window.LittledataLayer.attributes);
 	const attributesInCart = Object.keys(cart.attributes);
-	const allAttributesInCart = attributesToSet.every((attribute) => attributesInCart.includes(attribute));
```
lyykfi, line 77: I think you have to run 'npm run lint -- --fix'

lyykfi (changes requested):

src/common/helpers/getTrackingConsent.ts
```diff
@@ -0,0 +1,48 @@
...
+	return new Promise(resolve => {
+		const optanonWrapper = () => {
+			// we are looking for cookie category 2 (performance)
+			if (window.OptanonActiveGroups.includes(performanceCookieCategory)) resolve();
+		};
+		optanonWrapper();
+		// else wrap and trigger after cookie opt-in
+		window.OptanonWrapper = optanonWrapper;
+		// and ping our server to say the user hasn't opted in
+		setClientID(clientIdIfNotTracked, platform);
```
lyykfi, line 33: copy/paste setClientID(clientIdIfNotTracked, platform); in both function.
I think we should avoid it somehow.

lyykfi (changes requested):

src/common/helpers/getTrackingConsent.ts
edwardupton: I don't see how. It needs to be here, not in the parent function

lyykfi:
Looks nice!

Are we consider unit tests here?

edwardupton:
We should yes. I'll work on that later

PR #206, opened 2021-04-28
PD-413 Fix measurementId to MASTER

edwardupton:


src/gaTracker/helpers.ts
```diff
@@ -56,7 +56,7 @@ export const sendPageview = () => {
 	const page_location = cleanPageviewUrl(locationWithMedium);
 
 	if (hasGA4()) {
-		gtag('config', LittledataLayer.measurementID, {
+		gtag('config', LittledataLayer.measurementId, {
```
fstefan7, line 59: @edwardupton Won't the users be affected on the period between ld-news and shopify-tracker deploys? As shopify-tracker won't find the right id for a period of time until both the repo are synced.

Secondly, I think the users that have the script already inserted and update the shopify-tracker version won't have the `measurementId` field

edwardupton:
This only affects 4 stores on private beta so it's manageable. In most
cases the data layer already has measurementId, so this will instantly fix
it.

fstefan7 (approved):

PR #207, opened 2021-04-28
PD-413 Measurement id fix to STAGING

edwardupton:


fstefan7 (changes requested):
@edwardupton The `dist/` files should not be changed

PR #208, opened 2021-04-28
PD-413 Fix measurementId to STAGING

edwardupton:


index.d.ts
```diff
@@ -48,7 +48,7 @@ export interface OwnLayer {
 	hideBranding?: boolean;
 	writeKey?: string;
 	webPropertyID?: string;
```
fstefan7, line 50: @edwardupton Shouldn't we change all the fields that are in the same scenario? ex: `webPropertyID`

fstefan7 (approved):

index.d.ts
edwardupton: @fstefan7 that's a much bigger change, because it means updating the data layer for hundreds of merchants. I'll create another card.

edwardupton:
@fstefan7 please review again

fstefan7 (approved):

PR #209, opened 2021-04-29
PD-330 Get trackers

edwardupton:


src/gaTracker/helpers.ts
```diff
@@ -41,7 +43,8 @@ export const initGtag = () => {
 };
 
 const ga4EventParams = () => {
-	const trackers = window.ga && window.ga.getAll;
+	const trackers = window.ga && window.ga.getAll && window.ga.getAll();
```
lyykfi, line 46: window?.ga?.getAll?.()

lyykfi (changes requested):

src/gaTracker/helpers.ts
fstefan7: changed to use optional chaining

lyykfi (approved):

ionescurares (approved):

PR #210, opened 2021-04-30
PD-274 Store IP in local storage

edwardupton:


src/segmentTracker/helpers.ts
```diff
@@ -24,17 +25,22 @@ export const getContext = () => {
...
 			return resolve(standardContext);
 		}
+		const ipFromStorage = window.sessionStorage.getItem(sessionStorageKey);
+		if (ipFromStorage) {
+			return resolve({
+				...standardContext,
+				ip: ipFromStorage,
+			});
+		}
 		httpRequest
```
lyykfi, line 35: I would suggest to isolate the logic to a special helper like AnonymizedIpResolve for example.

lyykfi (changes requested):

src/segmentTracker/helpers.ts
```diff
@@ -176,3 +163,30 @@ export const callSegmentPage = (integrations: Record<string, any>) => {
...
+	return new Promise(resolve => {
+		const sessionStorageKey = 'littledata_ip_address';
+		const ipFromStorage = window.sessionStorage.getItem(sessionStorageKey);
+		if (ipFromStorage) {
+			return resolve({
+				...standardContext,
+				ip: ipFromStorage,
+			});
+		}
+		httpRequest
```
lyykfi, line 177: what do you think to use async/await here?

src/segmentTracker/helpers.ts
```diff
@@ -176,3 +163,30 @@ export const callSegmentPage = (integrations: Record<string, any>) => {
 		);
 	});
 };
+
+const anonymizedIpResolve = (standardContext: object) => {
```
lyykfi, line 167: I would rather suggest don't have standardContext here.
I guess anonymizedIpResolve  should return only IP which was resolved.

What do you think?

lyykfi (changes requested):

src/segmentTracker/helpers.ts
edwardupton: We've had bad experiences of using async/await with webpack. Best not to

edwardupton: Making it webpack compatible with async/await doubles the size of the minified code and makes debugging much harder

lyykfi: That is fine. Though, what was a problem?

MihaiCernat: yes, it messes up badly sometimes

lyykfi: Agree, let skip for now.

src/segmentTracker/helpers.ts
```diff
@@ -176,3 +163,24 @@ export const callSegmentPage = (integrations: Record<string, any>) => {
...
+		httpRequest
+			.getJSON('//api.ipify.org?format=json')
+			.then((response: any) => {
+				const { ip } = response;
+				const anonymizedIp = ip.replace(/([0-9a-z]+)$/, 0);
+				window.sessionStorage.setItem(sessionStorageKey, anonymizedIp);
+				resolve({ anonymizedIp });
+			})
+			.catch(() => {
+				resolve({});
```
lyykfi, line 183: should we use reject here?

MihaiCernat:
> should we have reject here?

@lyykfi no, some tests are failing, checking now

test/segmentTracker/getContext.test.ts
```diff
@@ -19,43 +19,41 @@ describe('getContext function', () => {
...
 		const context = await getContext();
-		//@ts-ignore
 		context.ip.should.equal('2001:0db8:85a3:0000:0000:8a2e:0370:0');
 	});
-	it('should call the right url to get the ip', async () => {
-		getJSON.resolves({ ip: '192.168.1.124' });
-		await getContext();
-		//@ts-ignore
-		getJSON.should.have.been.calledWith('//api.ipify.org?format=json');
+	it('does not call API if cached', async () => {
```
lyykfi, line 53: I think we should add a test for reject state.

lyykfi (changes requested):

lyykfi (approved):

PR #211, opened 2021-05-04
PD-366 uncaught promise

MihaiCernat:


src/segmentThankYouPageTracker/segmentThankYouPageTracker.ts
```diff
@@ -32,11 +32,15 @@ declare let window: CustomWindow;
...
-				userID: window.Shopify.checkout.customer_id,
-				segment: true,
-			});
+			httpRequest
+				.postJSON(clientIDEndpoint, {
+					clientID: user().anonymousId(),
+					userID: window.Shopify.checkout.customer_id,
+					segment: true,
+				})
+				.catch(() => {
```
lyykfi, line 41: Could we add a test for this case please?

lyykfi (changes requested):

lyykfi (approved):

PR #212, opened 2021-05-05
TF-6 Refactoring: remove ts-loader

lyykfi:


package.json
```diff
@@ -1,17 +1,17 @@
 {
 	"scripts": {
-		"build-master": "node ./node_modules/webpack/bin/webpack.js",
-		"build-staging": "node ./node_modules/webpack/bin/webpack.js --mode none",
-		"start": "node ./node_modules/webpack/bin/webpack.js --watch",
-		"lint": "node ./node_modules/eslint/bin/eslint.js --ext .ts ./src",
+		"build-master": "npx webpack --mode=production",
```
edwardupton, line 3: Is `mode=production` the same as `mode production`?

package.json
```diff
@@ -36,10 +37,9 @@
 		"prettier": "1.18.2",
 		"sinon": "2.4.1",
 		"sinon-chai": "2.12.0",
-		"terser-webpack-plugin": "5.1.1",
-		"ts-loader": "6.2.2",
+		"terser-webpack-plugin": "^5.1.1",
 		"ts-node": "9.1.1",
-		"typescript": "3.9.9",
+		"typescript": "^4.2.4",
```
edwardupton, line 42: Does eslint typescript plugin need updating too?

edwardupton (approved):

package.json
lyykfi: Should be the same.

package.json
lyykfi: Good point!

PR #239, opened 2021-05-25
PD-508 Gtm transaction event

edwardupton:


src/common/helpers/getOrderNumberFromHTML.ts
```diff
@@ -0,0 +1,10 @@
+export const getOrderNumberFromHTML = (document: HTMLDocument): string => {
+	const orderNumberHTML = document.getElementsByClassName('os-order-number')[0].innerHTML;
```
lyykfi, line 2: it would raise an error if document.getElementsByClassName('os-order-number').length === 0

test/gaThankYouPageTracker/shopifyCheckout.ts
```diff
@@ -0,0 +1,126 @@
+export default {
```
lyykfi, line 1: maybe should be shopifyCheckout.test.ts?

lyykfi (changes requested):

PR #241, opened 2021-06-09
PD-514 Fix/web property

edwardupton:


lyykfi (approved):

src/gaThankYouPageTracker/gaThankYouPageTracker.ts
```diff
@@ -22,16 +22,16 @@ declare let window: CustomWindow;
 		...convertParams(urlToQueryStringObject(scriptSrc)),
 	};
 	if (window.Shopify.Checkout && window.Shopify.Checkout.page === 'thank_you') {
-		const { webPropertyID, env } = window.LittledataLayer;
+		const { webPropertyId, env } = window.LittledataLayer;
```
lyykfi, line 25: should we use here getWebPropertyId() ?

lyykfi (changes requested):

src/gaThankYouPageTracker/gaThankYouPageTracker.ts
edwardupton: No we already handle it here:
https://github.com/littledata/shopify-tracker/pull/241/files#diff-ea1c824af3f4fa208df8722dbb61c4d6d6dfe2b352fb94a84eaa64a1672c350d

lyykfi (approved):

PR #252, opened 2021-06-14
PD-520 Trust-Arc integration

edwardupton:


lyykfi:
@edwardupton maybe we should change a target to merge to staging? also, i would suggest to change the default target from master to staging.

edwardupton:
@lyykfi No there is nothing to test on staging - this can only be tested once we have a link we can use with the customer, as we don't have access to TrustArc.

The alternative is to clone shopify-tracker repo into something like `ld-shopify-scripts` and push it only S3 from there.

src/customerPrivacyAPI/trustArc.ts
```diff
@@ -0,0 +1,72 @@
...
+	console: { (...data: any[]): void; (message?: any, ...optionalParams: any[]): void };
+	PrivacyManagerAPI: LooseObject;
+}
+
+declare let window: TrustArcWindow;
+
+(function () {
+	const COOKIE_CATEGORY = 2; // Cookies for Performance
+	let userCanBeTracked: boolean = window.Shopify?.customerPrivacy?.userCanBeTracked() || false;
+	const setTrackingConsent = (userCanBeTracked: boolean) =>
```
lyykfi, line 13: maybe just const setTrackingConsent = window.Shopify?.customerPrivacy?.setTrackingConsent; ?

src/customerPrivacyAPI/trustArc.ts
```diff
@@ -0,0 +1,72 @@
...
+    get's the new consent decision, and pushes the events into the GTM framework.
+    Once the event is submitted, that consent decision is marked in the 'dispatched' map so it does not occur more than once.
+   */
+	self.addEventListener('message', function (event) {
+		try {
+			let decision;
+
+			if (
+				event.data &&
+				(decision = JSON.parse(event.data)) &&
```
lyykfi, line 48: I would suggest decision = JSON.parse(event.data) before if because parse could throw an error.

lyykfi (changes requested):

src/customerPrivacyAPI/trustArc.ts
edwardupton: Can you clarify?

src/customerPrivacyAPI/trustArc.ts
edwardupton: No, because I want to set a blank callback in all cases

src/customerPrivacyAPI/trustArc.ts
```diff
@@ -0,0 +1,72 @@
...
+					userCanBeTracked = true;
+					setTrackingConsent(userCanBeTracked);
+				} else if (userCanBeTracked) {
+					console.log('UNTRACKED');
+					userCanBeTracked = false;
+					setTrackingConsent(userCanBeTracked);
+				}
+			}
+		} catch (xx) {
+			/** not a cm api message **/
```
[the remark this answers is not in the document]
edwardupton: No, this error means we shouldn't be handling the message

src/customerPrivacyAPI/trustArc.ts
lyykfi: Sure. if a JSON in event.data is wrong somehow this JSON.parse(event.data) will throw en error. should we wrap it by try/catch or we don't expect any problems there?

edwardupton:
The whole if() is in a try/catch, so unecesssary

lyykfi:
@edwardupton Looks good for me! However, should we add some tests here?

lyykfi:
@edwardupton Seems we don't have codeclimate connected here. Also, our codeship pipeline is poor. I'll create a card to code dept.

edwardupton:
Yes, we need test coverage for shopify-tracker.

For this particular feature, unit tests are hard because I can't find any
documentation on TrustArc's API

lyykfi:
@edwardupton That is fine. however, that would be nice at least to mock function and detect that it was actually calling. what do you think?

edwardupton:
> @edwardupton That is fine. however, that would be nice at least to mock function and detect that it was actually calling. what do you think?

@lyykfi I spent half an hour on this, but since it's not my code the structure is hard to test. I was blocked trying to mock window.postMessage, as the code uses self.top.postMessage, which can't be easily injected.

This was only supposed to be a quick fix for one customer. I've added one test, but LMK if you think we can easily add more

lyykfi:
@edwardupton that is fine. thanks.

could you fix changing a dist folder again pls?

edwardupton:
> @edwardupton that is fine. thanks.
> 
> could you fix changing a dist folder again pls?

@lyykfi changed PR to staging.

lyykfi (approved):

PR #265, opened 2021-06-25
PD-560 Use active currency

edwardupton:


src/gaTracker/helpers.ts
```diff
@@ -310,7 +310,7 @@ export const hasGA3 = (): boolean => !!getWebPropertyId();
 function convertProductsToGa4Format(products: Detail[], sendIndex: boolean): GA4Product[] {
 	return products.map((product) => {
 		const converted = {
-			currency: (LittledataLayer.ecommerce && LittledataLayer.ecommerce.currencyCode) || '',
+			currency: window.Shopify?.currency?.active || LittledataLayer.ecommerce?.currencyCode, //fallback to shop currency
```
lyykfi, line 313: i think we should use getActiveCurrency  here.

lyykfi (changes requested):

src/segmentTracker/helpers.ts
```diff
@@ -86,7 +86,6 @@ export const trackEvents = (): void => {
 				'Product Clicked',
 				{
 					...segmentProduct(product),
-					currency: LittledataLayer.ecommerce.currencyCode,
```
lyykfi, line 89: hm.... i'm not sure. should we also use getActiveCurrency(), here?

lyykfi: @edwardupton

edwardupton: I removed it because we already fetch the currency in `segmentProduct()` - it is redundant

edwardupton:
@lyykfi you are too fast! I was committing fixes for your issues as you were writing

lyykfi:
@edwardupton that is my style :-) Thank you for fixing. Looks good!

PR #282, opened 2021-07-19
PD-676 Fix for using Shopify customer privacy API before window.Shopify is ready

edwardupton:


src/segmentThankYouPageTracker/segmentThankYouPageTracker.ts
```diff
@@ -22,7 +22,7 @@ declare let window: CustomWindow;
 		ecommerce: {},
 		...convertParams(urlToQueryStringObject(scriptSrc)),
 	};
-	if (window.Shopify.Checkout && window.Shopify.Checkout.page === 'thank_you') {
+	if (window.Shopify?.Checkout?.page === 'thank_you') {
```
lyykfi, line 25: I think that is a good time to move thank_you to a dict. :-) Wdyt?

lyykfi (changes requested):

src/segmentThankYouPageTracker/segmentThankYouPageTracker.ts
edwardupton: done

lyykfi (approved):

PR #283, opened 2021-07-21
PD-702 Segment identify calls include `default_address` rather than `address`

edwardupton:


src/segmentTracker/helpers/identifyCustomer.ts
```diff
@@ -0,0 +1,47 @@
+import { CustomWindow } from '../../..';
+import { getCookie } from '../../common/getCookie';
+import { setCartOnlyAttributes } from '../../common/setClientID';
+import { addUserIdForCustomer } from './addUserIdForCustomer';
+
+declare let window: CustomWindow;
```
lyykfi, line 6: we do that in many places: `declare let window: CustomWindow;`
I think we could create a file main.d.ts and add the CustomWindow declaration there.

Only what we need todo after to add to package.json this string: "typings": "main.d.ts"

src/segmentTracker/helpers/identifyCustomer.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const identifyCustomer = (): void => {
+	const { customer, cookiesToTrack = [] } = window.LittledataLayer;
+
+	if (!customer) return;
+
+	const cookieTraits: any = {};
```
lyykfi, line 13: any  -> Record<string, string>

src/segmentTracker/helpers/identifyCustomer.ts
edwardupton: But isn't that what 'index.d.ts' does? Can you make the change on another PR as it's unrelated to this?

lyykfi (approved):

lyykfi:
@edwardupton could you fix conflicts pls?

PR #285, opened 2021-08-02
PD-631 - Facebook Pixel IDs

lageorgem:
* Added logic for extracting fbc and fbp cookies
* Added logic for storing said cookies into redis


src/segmentTracker/segmentTracker.ts
```diff
@@ -39,6 +39,31 @@ import { getTrackingConsent } from '../common/helpers/getTrackingConsent';
 					}
 				}
 			});
+
+			if (window?.fbq?.loaded) {
+				const cookies = document.cookie
```
lyykfi, line 44: we already have 'getCookie' helper.

index.d.ts
```diff
@@ -56,6 +56,7 @@ export interface CustomWindow extends Window {
 	_ga_originalSendHitTask: any;
 	OnetrustActiveGroups: string;
 	OptanonWrapper: CallableFunction;
+	fbq: any;
```
lyykfi, line 59: let keep it more specific.
fbq: {
loaded: boolean;
}

lyykfi (changes requested):

src/segmentTracker/segmentTracker.ts
lageorgem: Done

lageorgem: Done

index.d.ts
lageorgem: Done

src/segmentTracker/segmentTracker.ts
```diff
@@ -39,6 +40,19 @@ import { getTrackingConsent } from '../common/helpers/getTrackingConsent';
 					}
 				}
 			});
+
+			if (window?.fbq?.loaded) {
```
edwardupton, line 44: This should happen within window.analytics.ready callback, so we have a chance for FB library to add the cookie.

And can we import as a function like `retrieveAndStoreFacebookIds`? So this file stays clean

lyykfi: @edwardupton we are talking about the same thing. :-)

lyykfi: @lageorgem Just copy from a separated comment:

we do the same code twice here plus we don't isolate the FB logic from Segment.
I would suggest at least to add a helper file with functions to operate with FB or maybe a singletone class.

That could be like: FacebookCookieHelper.getItems(['_fbp', '_fbc'])

lageorgem: @edwardupton I will move the logic separately, but I don't think we should wait for GA to load in order to get the facebook cookies. It's two separate scripts which probably load asynchronously. Is there a reason for that call in the analytics ready callback?

edwardupton: @lageorgem this is not the callback for GA, but rather the callback for Segment.

We want to search for the cookies at the _latest time_ possible, to maximise the chance of FB library loading. There is no point sending fbc before the Segment library has loaded, because without the pageview the fbc value is useless.

Unfortunately there is no callback for `fbq`: https://stackoverflow.com/questions/29487133/facebook-conversion-event-callback

So I suggest it's better to fetch a bit later than earlier in the page load.

edwardupton: Ah actually the solution is in `waitForGAtoLoad` function - we can check for existence of `fbq` with exponential back off - similar to what you have done, but keep trying.

edwardupton: And yes, it doesn't need to be within analytics.ready callback.

lageorgem: It's still a race condition. I added a timeout that checks every 50ms for window.fbq.loaded until it's loaded, but if it never loads the timeouts continue to check infinitely. There should be a better way to approach this.

edwardupton: OK, it just needs unit tests then.

src/common/helpers.ts
```diff
@@ -153,3 +153,22 @@ export const getGAClientId = (tracker: LooseObject): string => {
...
 };
+
+export function retrieveAndStoreFacebookIds(): void {
+	if (!window?.fbq?.loaded) {
+		setTimeout(retrieveAndStoreFacebookIds, 50);
+
+		return;
+	}
+
+	const fbpClientId = getCookie('_fbp').split('.').splice(2, 3).join('.');
```
lyykfi, line 164: I would suggest to have a separated function to parse cookies for FB.

parseFBCookieByName = (name) => {
const fbcClientId = getCookie(name).split('.').splice(2, 3).join('.');
}

src/common/helpers.ts
```diff
@@ -153,3 +153,22 @@ export const getGAClientId = (tracker: LooseObject): string => {
 
 	return getValidGAClientId(clientId) ? clientId : '';
 };
+
+export function retrieveAndStoreFacebookIds(): void {
+	if (!window?.fbq?.loaded) {
```
lyykfi, line 158: could we have a generic function to track loading state like

const waitFor = (pattern, timeout = 50) => {
 return new Promise((resolve, reject) => {
   if (!pattern()) {
     setTimeout(waitFor , timeout );  
   } else {
    resolve()
   }
});
}

lyykfi (changes requested):

src/segmentTracker/segmentTracker.ts
lageorgem: Check it out

src/common/helpers.ts
lageorgem: yes

src/common/helpers.ts
lageorgem: Good old callback function solved it

lyykfi (approved):

PR #290, opened 2021-08-05
PD 661 - GA4 Session

lageorgem:
PD-661 GA4 Session ID

* Refactored waitForGaToLoad
* Saved GA measurementId cookie


src/common/helpers.ts
```diff
@@ -126,26 +123,68 @@ export const documentReady = (callback: CallableFunction): void => {
...
+ * from trackers.
+ * @param callback - Optional callback that will be executed on GA load.
+ */
+export const waitForGaToLoad = (callback?: CallableFunction): void => {
 	// After GA queue is executed we need to wait
 	// until after ga.getAll is available but before hit is sent
-	const trackers = window.ga && window.ga.getAll && window.ga.getAll();
-
-	if (trackers && trackers.length) {
+	const checkCallback: CheckCallback = () => {
```
lyykfi, line 137: That function can be moved from waitForGaToLoad because there isn't any dependencies. Also, it will fix codeclimate complains.

src/common/helpers.ts
```diff
@@ -126,26 +123,68 @@ export const documentReady = (callback: CallableFunction): void => {
...
+		const trackers = window.ga && window.ga.getAll && window.ga.getAll();
+
+		if (trackers && trackers.length) return trackers;
+		else return false;
+	};
+	const foundCallback: FoundCallback = (err: Error, trackers: any): void => {
```
lyykfi, line 143: the same.

src/segmentTracker/helpers/retrieveAndStoreFacebookIds.ts
```diff
@@ -1,32 +1,13 @@
...
-
-		return;
-	}
-
-	callback(null);
-}
-
 export function retrieveAndStoreFacebookIds(): void {
-	onFbLoaded((err: Error) => {
+	const checkCallback: CheckCallback = () => window?.fbq?.loaded;
```
lyykfi, line 9: move out from the function.

src/segmentTracker/helpers/retrieveAndStoreGA4Id.ts
```diff
@@ -0,0 +1,23 @@
...
+
+declare let window: CustomWindow;
+
+function parseGA4Cookie(cookieName: string) {
+	return getCookie(cookieName).split('.').splice(2, 1).join();
+}
+
+export function retrieveAndStoreGA4Id(): void {
+	const foundCallback: CallableFunction = () => {
+		const [_, measurementId] = window.LittledataLayer.measurementId.split('-');
```
lyykfi, line 14: I am not sure but I guess you can write just [, measurementId]

lyykfi (changes requested):

src/common/helpers.ts
lageorgem: I did it because I found it to make more sense being inside the scope of the waitForGaToLoad. It would definetly work outside that function also, but i think it's harder to read this way.

lyykfi: actually that could be a separated function with name 'getGATrackers()'. logically the function does separated thing.

src/common/helpers.ts
lageorgem: I'd rather keep the foundCallback inside the function, because it uses the callback provided as a parameter in the upper scope, and I find it cleaner to use that instead of passing it along.

lyykfi (approved):

ionescurares (approved):

PR #292, opened 2021-08-10
PD-766 Post attributes to cart only once for stores running multiple tracking script

edwardupton:


src/common/setClientID.ts
```diff
@@ -23,10 +23,15 @@ export const setClientID = (clientId: string, platform: platformTypes): void =>
 	window.LittledataLayer.attributes[clientIDProperty] = clientId;
 
 	clearTimeout(postCartTimeout);
-	// timeout is to allow 2 client IDs posted within 1 second
+	// timeout allows multiple client IDs posted within 1 second
 	// to be included in the same cart update
 	postCartTimeout = setTimeout(() => {
```
lyykfi, line 28: do we need still the timeout it this case?

edwardupton: yes even more so - this allows us to send all calls to `setClientId` from any async function, within 1 second of the first call, in a single POST to Shopify and Littledata

lyykfi (approved):

PR #298, opened 2021-08-11
PD-783 Product Tags Segment

lageorgem:
* Captured tags from shopify and sending them into segment


lyykfi:
@lageorgem could you fix tests pls?

index.d.ts
```diff
@@ -14,7 +14,7 @@ export interface Detail {
 	shopify_variant_id?: string;
 	compare_at_price?: string;
 	image_url?: string;
-	tags?: string;
+	tags: string;
```
edwardupton, line 17: This interface is reused by GA events, where tags would not be sent - so needs to be optional.

Or we could extend the interface to `SegmentDetail` and include extra Segment fields in that.

edwardupton (changes requested):

PR #306, opened 2021-09-08
PD-864 More properties for Segment "Cart Viewed" and "Product Searched" events

edwardupton:


src/segmentThankYouPageTracker/helpers/mapSegmentProducts.ts
```diff
@@ -0,0 +1,13 @@
+import { SegmentProduct } from '../../../segmentInterface';
+
+export const mapSegmentProducts = (shopifyProducts: LooseObject, pricesIsInCents?: boolean): [SegmentProduct] => {
```
lyykfi, line 3: We are going to write comment for functions to enhance code readability.
Would be nice if you could write down a small comment for the function with a short props description

edwardupton: See comment

lyykfi (approved):

PR #307, opened 2021-09-14
PD-873 Add cookiesToTrack to cart attributes even before a customer is logged in

edwardupton:


edwardupton:
@lyykfi do we need to add Cypress tests to CI pipeline?

lyykfi:
@edwardupton I think we should have. Every time when you merge something to staging Cypress should be triggered.

PR #310, opened 2021-09-23
PD-864 Wait for search results before sending Products Searched

edwardupton:


src/segmentTracker/helpers/sendEventsWithPageview.ts
```diff
@@ -19,9 +19,23 @@ export const sendEventsWithPageview = (pathname: string): void => {
 		trackEvent('Registration Viewed', {});
 	}
 	if (pathname === '/search') {
+		trackSearch();
+	}
+};
+
+function trackSearch(secondAttempt?: boolean): void {
```
lyykfi, line 26: could you add a comment please?

src/segmentTracker/helpers/sendEventsWithPageview.ts
```diff
@@ -19,9 +19,23 @@ export const sendEventsWithPageview = (pathname: string): void => {
...
+	const results = getProductLinks(document, LittledataLayer.productListLinksHaveImages).length;
+
+	if (results == 0 || secondAttempt) {
```
lyykfi, line 29: can be just !results || secondAttempt

src/segmentTracker/helpers/sendEventsWithPageview.ts
```diff
@@ -19,9 +19,23 @@ export const sendEventsWithPageview = (pathname: string): void => {
...
 			query: document.location.search.replace('?q=', '').replace('+', ' '),
-			results: getProductLinks(document, LittledataLayer.productListLinksHaveImages).length,
+			results,
 		});
+
+		return;
 	}
-};
+	// check again in case the products list took a long time to load
+	window.setTimeout(() => {
```
lyykfi, line 38: should we try only twice or maybe many times?

edwardupton:
Added comment.

I think just twice, because if there really are no products we do want to
track the event with zero products

lyykfi (approved):

PR #320, opened 2021-12-16
product_delay

ionescurares:


src/common/productListViews.ts
```diff
@@ -11,7 +11,11 @@ require('swiped-events');
 type impressionCallback = (impressionTag: Impression[]) => void;
 
 const impressionsToSend = [] as ImpressionToSend[];
-const millisecondsForProductAPI = 800;
+let millisecondsForProductAPI = 800;
```
TheGame2500, line 14: Could just be const millisecondsForProductApi = Littledata.productDelay || 800

TheGame2500 (changes requested):

TheGame2500 (approved):

PR #321, opened 2022-01-07
PD-1069: context should be on all client side events and include externalids

andreivaleanu:
Added externalids to context and tested all the shopify client events that are sent to segment.


src/segmentTracker/helpers.ts
```diff
@@ -17,6 +17,8 @@ const getOnetrustActiveGroups = () => {
 };
 
 export const getContext = (): Promise<LooseObject> => {
+	const userId = window.LittledataLayer.customer.id;
```
fstefan7, line 20: @andreivaleanu This could fail if the `customer` is not set on the `LittledataLayer` object

src/segmentTracker/helpers.ts
```diff
@@ -38,6 +41,21 @@ export const getContext = (): Promise<LooseObject> => {
 	});
 };
 
+function getExternalIds(uid: string) {
+	if (!uid) return undefined;
```
fstefan7, line 45: I would change this to 
```
if (!uid) return;
```
as it would be the same returned value, but it's up to you on how you prefer

fstefan7 (changes requested):

fstefan7:
Please also add some unit tests checking for the context

src/segmentTracker/helpers.ts
```diff
@@ -17,6 +17,8 @@ const getOnetrustActiveGroups = () => {
 };
 
 export const getContext = (): Promise<LooseObject> => {
+	const userId = window.LittledataLayer?.customer?.id || undefined;
```
fstefan7, line 20: You can replace it with `const userId = window.LittledataLayer?.customer?.id`
When no `LittledataLayer`, or no `LittledataLayer?.customer` will be defined, then the `userId` will be undefined

andreivaleanu: yes. i know that it is redundant but i find it more easy to read. Will change.

fstefan7 (approved):
looks good to me 👍

PR #324, opened 2022-01-17
DPD-17 add thank you page client events context and identify hits context

andreivaleanu:
Add externalids to events that were missing it.
thank you page 
identify


src/segmentTracker/helpers.ts
```diff
@@ -17,7 +17,16 @@ const getOnetrustActiveGroups = () => {
 };
 
 export const getContext = (): Promise<LooseObject> => {
-	const userId = window.LittledataLayer?.customer?.id;
+	let userId = window.LittledataLayer?.customer?.id;
+
+	const { user } = window.analytics;
+
+	if (!userId) {
+		userId = window.Shopify.checkout.customer_id;
```
edwardupton, line 25: Needs conditional operators. `Shopify.checkout` won't be defined outside of checkout

src/segmentTracker/helpers.ts
```diff
@@ -17,7 +17,16 @@ const getOnetrustActiveGroups = () => {
...
+	}
+	if (!userId && typeof user === 'function') {
+		userId = user().id();
```
edwardupton, line 28: Are you sure `analytics.user().id()` is not the anonymous ID?

src/segmentTracker/helpers.ts
```diff
@@ -17,7 +17,16 @@ const getOnetrustActiveGroups = () => {
 };
 
 export const getContext = (): Promise<LooseObject> => {
-	const userId = window.LittledataLayer?.customer?.id;
+	let userId = window.LittledataLayer?.customer?.id;
```
edwardupton, line 20: Can you call it `shopifyCustomerId` for clarity? `userId` in the context of Segment could be another identifier: https://segment.com/docs/connections/sources/catalog/libraries/website/shopify-littledata/#user-identity

edwardupton (changes requested):

src/segmentTracker/helpers.ts
edwardupton: Either way let's drop this, as the Segment userId may not be the Shopify customer Id.

andreivaleanu: > Are you sure `analytics.user().id()` is not the anonymous ID?

it's not, i've checked.

andreivaleanu: > Either way let's drop this, as the Segment userId may not be the Shopify customer Id.

ok. will make the changes now.

edwardupton (approved):

PR #328, opened 2022-02-04
hotfix DPD-58: Switch back to call page before analytics.ready()

edwardupton:


TheGame2500 (approved):
@edwardupton @andreivaleanu this looks fine but we should really start using async/await in this repo instead of the weird Promise/callback mix we do to make order of execution clearer.

What i mean is that `retrieveAndStoreClientId` will roughly, under the hood, wait for window.analytics.ready (GA ready) to fire and run at about the same time. But to understand that i had to go through a lot of code + conversation.

edwardupton:
@TheGame2500 we did test using async, but transpiling it back into ES2015 resulted in a 50% bigger minified file. Maybe worth testing again, but it is a big job.

ionescurares (approved):

TheGame2500:
@edwardupton yes I remember the files being bigger but it seems like it's worth it for code clarity. @andreivaleanu for now, can you actually add some comments as to why those lines are in the particular order they are in. Perhaps some links to this bug would be useful

PR #336, opened 2022-03-16
DPD-142 Fix/dpd 142 ga4 inconsistent master forked

igorkol:


src/gaTracker/gaTracker.ts
```diff
@@ -20,12 +20,11 @@ declare let window: CustomWindow;
 		advertiseLD('Google Analytics');
 		retrieveAndStoreClientId();
 		retrieveAndStoreGA4Id();
-		documentReady(() => {
-			trackEvents();
-		});
+		documentReady(() => {});
```
fstefan7, line 23: @igorkol I don't think we still need the `documentReady` here, as its callback does nothing after moving the `trackEvents` method inside the pageView

src/gaTracker/helpers.ts
```diff
@@ -308,17 +335,23 @@ export const hasGA3 = (): boolean => !!getWebPropertyId();
 
 function convertProductsToGa4Format(products: Detail[], sendIndex: boolean): GA4Product[] {
 	return products.map((product) => {
+		let convPrice: number = +product.compare_at_price;
```
fstefan7, line 338: I suggest renaming the variables to something as `price`, `discountedPrice`, `discount`

src/gaTracker/helpers.ts
```diff
@@ -246,8 +269,11 @@ function sendViewItemListEvent(products: Impression[]): void {
 
 function sendViewItemEvent(product: Detail): void {
 	if (hasGA4()) {
+		let price: number = +product.price;
 		gtag('event', 'view_item', {
 			...ga4EventParams(),
+			currency: getActiveCurrency(),
+			value: price,
```
fstefan7, line 276: I think I wouldn't use a variable here, since we're only using it once.

If really wanting to have the product.price in a variable, you could define it as `const` in order to block it from being able to be modified in the future.

Also, you can have it like:
```
gtag('event', 'view_item', {
  ...ga4EventParams(),
  currency: getActiveCurrency(),
  value: +product.price,
```

src/gaTracker/helpers.ts
```diff
@@ -16,28 +16,39 @@ const getWebPropertyId = () => LittledataLayer.webPropertyId || LittledataLayer.
...
 		};
 	window.ga.l = +new Date();
 
+	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
 	// @ts-ignore
 	gtag('js', new Date());
 
-	if (hasGA4()) {
-		gtag('config', LittledataLayer.measurementId, {
+	if (hasGA3() && hasGA4()) {
```
fstefan7, line 38: For now it would work even with only this if statement 
```
if (hasGA3()) {
 ...
}
```

but up to you to keep as is if you think we'll have soon the possibility of having GA4 without GA3 and if it would help you then

fstefan7 (changes requested):
Looks good, left some minor comments 👍

src/gaTracker/helpers.ts
igorkol: done

src/gaTracker/helpers.ts
igorkol: done

src/gaTracker/gaTracker.ts
igorkol: done

src/gaTracker/helpers.ts
```diff
@@ -269,11 +278,10 @@ function sendViewItemListEvent(products: Impression[]): void {
 
 function sendViewItemEvent(product: Detail): void {
 	if (hasGA4()) {
-		let price: number = +product.price;
 		gtag('event', 'view_item', {
 			...ga4EventParams(),
 			currency: getActiveCurrency(),
-			value: price,
+			value: <number>+product.price,
```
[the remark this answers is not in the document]
igorkol: that is intentional, to force it being a number and so that no one changes the format in the future by mistake

PR #341, opened 2022-03-23
fix(dpd-143): add fbclid to /store even when fbq is not present

andreivaleanu:


src/segmentTracker/segmentTracker.ts
```diff
@@ -35,7 +36,24 @@ declare let window: CustomWindow;
...
+				if (!window.LittledataLayer.FacebookPixelID) {
+					if (window.analytics.Integrations['Facebook Pixel']) {
+						//if segment has facebook destination retrieve and store fbc and fbp from cookies
+						retrieveAndStoreFacebookIds();
+					} else {
+						//if segment does not have facebook destination save fbc from query url if present
+						const queryObj = urlToQueryStringObject(window.location.search);
+
+						if (queryObj.fbclid) {
+							const fbcClientId = `fb.2.${Date.now()}.${queryObj.fbclid}`;
```
igorkol, line 48: I assume you used this FB spec for it: https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/fbp-and-fbc/
Based on what they say, if `fbc` is built on the server-side and `_fbc` is not present, `1` should be used as the domain identifier.
So, I think you should change this line to:
const fbcClientId = `fb.1.${Date.now()}.${queryObj.fbclid}`;

igorkol (changes requested):
Just one confirmation and alignment with FB docs.
Otherwise, changes are fine

PR #348, opened 2022-04-14
Fix/dpd 142 ga4 inconsistent master forked

igorkol:


src/common/setClientID.ts
```diff
@@ -7,7 +7,7 @@ declare let window: CustomWindow;
 let postCartTimeout: any;
 const cartOnlyAttributes: LooseObject = {};
 
-export const setClientID = (clientId: string, platform: platformIdentifiers): void => {
+export const setClientID = (clientId: string | Promise<string | void>, platform: platformIdentifiers): void => {
```
fstefan7, line 10: We should be more strict here and only accept the clientId as a string. We can wait for the clientId promise to fulfil  before passing it to the function(even if I did not see where it is being passed as a promise).

Other ways it would be a pain in the ass to check inside the method what the clientId type is and so on, which would make the codebase worse and harder to debug/understand

src/segmentTracker/helpers/retrieveAndStoreGA4Id.ts
```diff
@@ -5,21 +5,30 @@ import { waitForGaToLoad } from '../../common/helpers';
 
 declare let window: CustomWindow;
 
-function parseGA4Cookie(cookieName: string) {
-	return getCookie(cookieName).split('.').splice(2, 1).join();
+function parseGA4Cookie(cookieName: string): Promise<string> {
+	return new Promise((resolve) => {
```
fstefan7, line 9: If the cookie is not present, then we're running it once every 50ms non stop. This could affect the response time of the store.

I suggest doing it as it's done on the GA3 clientId retrieval(a exponential time increase and with a specific timeout - as after X retries to stop). 
If we're sure that the cookie should be present and we want to retrieve it really fast, then your method could also be ok, but I would not let it as is, without a timeout when to stop if the cookie is not found

fstefan7 (changes requested):

src/common/setClientID.ts
igorkol: this was a remainder from one of the previous versions of the code that i forgot to clean up. yes, it doesn't need to be a promise and a string is enough.
thank you for catching this @fstefan7

src/segmentTracker/helpers/retrieveAndStoreGA4Id.ts
igorkol: good call - resolved

src/segmentTracker/helpers/retrieveAndStoreGA4Id.ts
```diff
@@ -5,21 +5,30 @@ import { waitForGaToLoad } from '../../common/helpers';
 
 declare let window: CustomWindow;
 
-function parseGA4Cookie(cookieName: string) {
-	return getCookie(cookieName).split('.').splice(2, 1).join();
+function parseGA4Cookie(cookieName: string): Promise<string | Error> {
+	return new Promise((resolve) => {
+		function repeat(count: number) {
```
fstefan7, line 10: When needed to throw an error there is no need to encapsulate it into a promise, the method can be as:

```
function parseGA4Cookie(cookieName: string): Promise<string> {
    if (isError) throw new Error('this is an error'); // not encapsulated into promise

   const cookie = await ga4.getCookie(cookieName);
   return cookie;
}
```

Also, the resolve callback returns a promise that succeeded, not a failed one, so in your case it would return a promise that finished properly with a successful result which in your case would have been an error.
In order to return a failed promise you should use the `reject` callback which is the second parameter of the Promise:
```
new Promise((resolve, reject)
```

src/common/getConfig.ts
```diff
@@ -15,7 +15,7 @@ export const DEFAULT_LINKER_DOMAINS = [
 
 export const extraExcludedReferrers = ['shop.app'];
 
-export default (): Gtag.CustomParams => {
+export default (ga3checker?: 'ga3'): Gtag.CustomParams => {
```
fstefan7, line 18: Here I suggest not using strings, but rather enums or booleans. You can define an enum as:
```
enum EventType {
 Ga3: 'ga3',
 Ga4: 'ga4'
}
```

and pass it as parameter as:

```
export default (ga3checker?: EventType.Ga3): Gtag.CustomParams => {
```

fstefan7 (changes requested):

src/segmentTracker/helpers/retrieveAndStoreGA4Id.ts
igorkol: as we discussed on the call, the purpose was to output an error to the console in case the counter runs out. i will wrap it up in the ld.debug check and change it to console.error and it will be okay

src/common/getConfig.ts
```diff
@@ -15,9 +15,9 @@ export const DEFAULT_LINKER_DOMAINS = [
 
 export const extraExcludedReferrers = ['shop.app'];
 
-export default (ga3checker?: 'ga3'): Gtag.CustomParams => {
+export default (ga3checker?: keyof typeof ConfigType): Gtag.CustomParams => {
```
fstefan7, line 18: If we're writing this way:
```
ga3checker?: keyof typeof ConfigType
```
then throughout the codebase, where we're calling this method we need to query it with a string(`getConfig('GA3')`) which somehow defies the purpose of using an enum. 

One improvement here is to have the type as 
```
ga3checker?: ConfigType
```

and everywhere we have `ga3` inside the code we can convert it to `ConfigType.GA3`


You can have a look at `HTTPMethod` enum which is used with the pattern that I raised above.

igorkol: as we discussed on slack, this solution above won't work like that because of a collision with another node library. i made a predefined list of strings instead of enum

fstefan7 (approved):
Looks good 👍

PR #350, opened 2022-04-21
DPD-169 Gads transaction tracking

igorkol:


src/common/helpers/urlToQueryStringObject.ts
```diff
@@ -35,11 +35,27 @@ const parseURLParams = (inputObject: LooseObject): LooseObject => {
...
+		if (sanitisedVal === 'true' || sanitisedVal === 'false') {
+			const mappedVal = getValFromBooleanMapping(sanitisedVal);
+
+			return {
+				...acc,
+				[paramName]: mappedVal,
+			};
+		}
+
+		if (paramName.includes('[')) {
```
fstefan7, line 47: I suggest adding some documentation here to explain its purpose

src/common/sendToGAds.ts
```diff
@@ -0,0 +1,48 @@
...
+				);
+			}
+		});
+	}
+
+	return;
+};
+
+export const sendAdsTransactionConversion = (adsIds: string[], value: number, currency: string): LooseObject => {
+	if (typeof adsIds === 'object' && adsIds.length > 0) {
```
fstefan7, line 22: Hmm, here you're checking that `adsIds` is an object and bellow you're doing `adsIds.forEach()` on it. The js object does not have a `forEach` or a `length` method

src/common/sendToGAds.ts
```diff
@@ -0,0 +1,48 @@
...
+		adsIds.forEach((id) => {
+			// checking if format is AW-123123/kadkfha_ab123
+			if (/^AW-[0-9]+(?:[\/][a-zA-Z0-9_-]+)$/.test(id)) {
+				return gtag('event', 'conversion', {
+					...prepareAdsConversion(id, value, currency),
+				});
+			} else {
+				console.error(`${id} ID doesn't contain a Google Ads conversion label. Conversion event was not sent`);
```
fstefan7, line 30: Do we want users to be able to see this error?

test/common/sendToGAds.test.ts
```diff
@@ -0,0 +1,71 @@
...
+				send_to: 'AW-123/adfasdf',
+				value: 25,
+				currency: 'USD',
+			});
+		});
+		it('regex should reject all the faulty conversion ids', () => {
+			const faultyIds = ['AW-123123', 'AB-123123/aljaldjflkj', 'aldjfaljdf'];
+
+			faultyIds.forEach((id) => {
+				expect(/^AW-[0-9]+(?:[\/][a-zA-Z0-9_-]+)$/.test(id)).to.be.false;
```
fstefan7, line 31: This should be encapsulated in a constant and used in both of the places. If not, then, in the future someone could change the regex from the code without changing it here, and the tests will still pass even if the regex is broken

fstefan7 (changes requested):
Added some minor comments

src/common/sendToGAds.ts
igorkol: well, js array is a primitive object type :) and array does have length and forEach.
however, now I realise that something else might sneak in here during the runtime. so, doing a Array.isArray([1, 2, 3]) check is a better idea 👍

src/common/sendToGAds.ts
igorkol: yes, we do, that is the purpose. this message is for them or their developers

src/common/helpers/urlToQueryStringObject.ts
igorkol: done

test/common/sendToGAds.test.ts
igorkol: done

PR #353, opened 2022-05-09
DPD-169 GAds transaction tracking

igorkol:
patched bug with two scenarios:
1. in which AW ID was set before and then removed. Ads pageviews would still be sent with an error because LD would contain "googleAdsConversionIds":[] and truthy condition would be met
2. trying to send conversion event on the ty page even if AW ID (with or without label) was never set. There was no check if the ID is set. That would output an error in the console.


src/gaTracker/helpers.ts
```diff
@@ -82,7 +82,7 @@ export const sendPageview = (): void => {
 	});
 
 	// sending Google Ads pageview, if AW Id is set
-	if (LittledataLayer.googleAdsConversionIds) sendAdsPageview(LittledataLayer.googleAdsConversionIds);
+	if (LittledataLayer.googleAdsConversionIds.length > 0) sendAdsPageview(LittledataLayer.googleAdsConversionIds);
```
fstefan7, line 85: This would lead to errors, if `LittledataLayer.googleAdsConversionIds` is undefined. You can change it to 
```
LittledataLayer.googleAdsConversionIds?.length
```

src/common/sendToGAds.ts
```diff
@@ -1,52 +1,96 @@
...
+
+	return;
+};
+
+/**
+ * Checks if the AW ID contains label and is in the right format
+ * @param {string} fullId AW ID and label (eg.'AW-123/dlkfajdlsj')
+ * @returns correct AW conversion ID and label or undefined
+ */
+const confirmFullFormat = (fullId: string): GAds.fullAwId | undefined => {
```
fstefan7, line 71: maybe a better naming would be `hasFullFormat` or `checkFullFormat`

fstefan7 (approved):

fstefan7:
I approved it, but there still is a conflict to fix

igorkol:
resolved conflict

PR #354, opened 2022-05-13
DPD-217 Hashed emails added to all dL events on frontend

igorkol:


igorkol:
codeclimate reports long code blocks. we have agreed already that the datalayer pushes should be stripped out of the same method and refactored into its own method. this will make writing the unit tests possible and also resolve the codeclimate warning

test/gaTracker/helpers.test.ts
```diff
@@ -0,0 +1,27 @@
...
+const fakePageURL = '/test';
+
+describe('gaHelpers', () => {
+	before(() => {
+		const gtagStub = '';
+
+		global.gtag = sinon.stub(null, gtagStub);
+	});
+	context('checking the pageview', () => {
+		it('should send the email hashed to dL if customer object available');
```
fstefan7, line 25: These are the unit tests you were saying that you would build on some other PR?

src/common/helpers/hashString.ts
```diff
@@ -0,0 +1,8 @@
+import MD5 from 'crypto-js/md5';
+
+/**
+ * Hashes (MD5) provided string
+ * @param string input string
+ * @returns md5 hashed string
+ */
+export const hashString = (string: string): string => (typeof string == 'string' ? MD5(string).toString() : undefined);
```
fstefan7, line 8: I suggest to not use keywords as `string` for variables.
Also, keeping in mind that you have the parameter already defined as `string` I suggest to not check if it is a string anymore.

The method could look like:

```
export const hashString = (stringToBeHashed: string): string => MD5(stringToBeHashed).toString();
```

fstefan7 (changes requested):

test/gaTracker/helpers.test.ts
igorkol: yes, this is a placeholder. it outputs the "pending" state in mocha

igorkol: but more would be created. this is only the first

src/common/helpers/hashString.ts
igorkol: I am checking the typeof because I want to prevent getting the `undefined` as an input. TS won't allow us to put anything else in there because of the type definition. But at runtime, we don't know what might go through as a value of an email :) Do you agree @fstefan7 ?

igorkol: Especially because this is a helper function. Now I know that almost for sure the email field won't go empty from Shopify customer object. But I don't know what this fn might be used for afterwards as well

fstefan7: I would say to not test inside the `hashString` method that a parameter was passed properly, but instead checking it as you did, when passing the argument to the method. 
I am fine anyway with both of the solutions 👍

fstefan7 (approved):

PR #359, opened 2022-05-31
DPD-218: Tracker to support Dawn 2.0

andreivaleanu:
Tested existing functionality to check events are being sent on the new theme.

Fix product list viewed event not firing because of html structure.


fstefan7 (approved):
LGTM!

PR #374, opened 2022-06-24
FR-32 DPD-195 UTMs in segment payload

igorkol:


src/common/constants.ts
```diff
@@ -18,3 +18,4 @@ export const localStorageNames = {
 	list_name: 'littledata_list_name',
 };
 export const AW_CONVERSION_REGEX = '^AW-[0-9]+(?:[/][a-zA-Z0-9_-]+)$';
+export const UTM_ORDER_IN_ATTRIBUTES = ['utm_source', 'utm_medium', 'utm_campaign', 'utm_content', 'utm_term'];
```
fstefan7, line 21: I don't understand very well the name of this array, are we looking that the utm parameters to be in this order?

src/common/helpers/removeObjectProperties.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Removing properties from an object based on a value
+ * @param removeValue - remove properties that contain this value
+ * @param inputObj - input object
+ * @returns output object without key/values that needed to be removed
+ */
+export const removeObjectPropertiesBasedOnValue = (removeValue: string, inputObj: LooseObject): LooseObject => {
```
fstefan7, line 7: Please add unit tests for this method as well

src/common/helpers/retrieveAndStoreUTMs.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Checks if UTM parameters exist and returns a UTM array
+ * @param fullQueryObj - object containing all the query params as keys
+ * @returns new array containing only UTM parameters, empty object if there are no UTMs,
+ * or undefined if there are no query params at all
+ */
+const checkIfUtmsExist = (fullQueryObj: LooseObject): undefined | string[] => {
+	// check if any query param exists
+	if (Object.keys(fullQueryObj).length === 0) return;
```
fstefan7, line 16: This is not needed, as 
```return Object.keys(fullQueryObj).filter((key) => key.includes('utm_'))```
treats this case as well

File: src/common/helpers/retrieveAndStoreUTMs.ts [hunk 4]
```diff
@@ -0,0 +1,66 @@
...
+
+declare let window: CustomWindow;
+
+/**
+ * Checks if UTM parameters exist and returns a UTM array
+ * @param fullQueryObj - object containing all the query params as keys
+ * @returns new array containing only UTM parameters, empty object if there are no UTMs,
+ * or undefined if there are no query params at all
+ */
+const checkIfUtmsExist = (fullQueryObj: LooseObject): undefined | string[] => {
```
fstefan7, line 14: Here the name of the function is `checkIfUtmsExist` which in my mind(from its name) would throw an error if utms are not present, but instead it returns an array of utm keys. So it does not really check if utms exists

I suggest to change the name to something like `hasUtmParameters` and to return a boolean, true if it has any utm parameters and false otherwise

src/common/helpers/retrieveAndStoreUTMs.ts
```diff
@@ -0,0 +1,66 @@
...
+
+	return Object.keys(fullQueryObj).filter((key) => key.includes('utm_'));
+};
+
+/**
+ * Get list of UTMs and convert back to object
+ * @param utmObj - object containing all the query params as keys
+ * @returns object of only UTM parameters with values
+ */
+const retrieveUTMs = (utmObj: LooseObject): undefined | LooseObject => {
```
fstefan7, line 26: Please add unit tests to the newly created methods

src/common/helpers/retrieveAndStoreUTMs.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @returns array of UTM parameter values
+ */
+export const formUTMString = (utms: LooseObject): string[] => {
+	const arrayOfEntries: string[] = [];
+
+	// we are pushing 'empty' as a placeholder of a position
+	for (const key of UTM_ORDER_IN_ATTRIBUTES) {
+		if (!utms[key]) {
+			arrayOfEntries.push('empty');
+			continue;
```
fstefan7, line 47: Instead of using `continue` you can add `arrayOfEntries.push(utms[key]);` inside an else branch. It would make the code a bit easier to read

src/segmentTracker/helpers/addUTMToEvents.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Reverting UTMs from string to an object and pushing to Segment's properties object
+ * @param properties Segment's properties object
+ * @returns Properties payload with the UTMs
+ */
+export const addUTMToProperties = (properties: LooseObject): LooseObject => {
+	const utmString = window.LittledataLayer?.attributes?.['UTM-clientID'];
+
+	if (!utmString) {
+		return { ...properties };
```
fstefan7, line 32: This is the same as `return properties`

src/common/helpers/retrieveAndStoreUTMs.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @param utms - object containing UTMs
+ * @returns array of UTM parameter values
+ */
+export const formUTMString = (utms: LooseObject): string[] => {
+	const arrayOfEntries: string[] = [];
+
+	// we are pushing 'empty' as a placeholder of a position
+	for (const key of UTM_ORDER_IN_ATTRIBUTES) {
+		if (!utms[key]) {
+			arrayOfEntries.push('empty');
```
fstefan7, line 46: I suggest pushing a falsy value, as `false` or `undefined`. In this case you can filter the object more easily with something as `[].filter(value => !!value)`

fstefan7 (changes requested):
Looks good, left a couple of comments.

src/common/constants.ts
igorkol: yes. i added the clarification in the filed

src/common/helpers/removeObjectProperties.ts
igorkol: done

src/common/helpers/retrieveAndStoreUTMs.ts
```diff
@@ -0,0 +1,65 @@
...
+export const formUTMString = (utms: LooseObject): string[] => {
+	const arrayOfEntries: string[] = [];
+
+	// we are pushing undefined as a placeholder of a position
+	for (const key of UTM_ORDER_IN_ATTRIBUTES) {
+		if (!utms[key]) {
+			arrayOfEntries.push(undefined);
+		} else {
+			arrayOfEntries.push(utms[key]);
+		}
```
fstefan7, line 46: This if - else branch could be converted to:
```
const utmParameter = utms[key] || undefined; // the undefined here is not mandatory, only for being easier to understand
arrayOfEntries.push(utmParameter);
```

if you find it more readable

fstefan7 (approved):
Looks nice! 👍

PR #375, opened 2022-06-29
DPD-248 | Add index.ts that dynamically imports and loads required scripts

Getriax:


src/common/helpers.ts
```diff
@@ -80,11 +80,26 @@ export const trackSocialShares = (clickTag: (name?: string) => void): void => {
 	});
 };
 
-export const validateLittledataLayer = (): void => {
+const wait = (milliseconds = 1000) => new Promise((resolve) => setTimeout(resolve, milliseconds));
```
igorkol, line 83: this is for LD layer that the Shopify theme will deliver, right? it might not load during that short time on the slower connections. we already have a mechanism with callbacks that waits for 60sec and checks every 50ms.
maybe have some callback mechanism instead that would try for bundled settings at least a few times?

igorkol:
no new tests @Getriax ?

igorkol:
also, no documentation/jsdoc for the new stuff? we are trying to add documentation to the new features we add to the tracker and to the older functions whenever we have to refactor

src/common/helpers.ts
Getriax: Sure, can do that

igorkol (approved):
all good

PR #377, opened 2022-07-10
DPD-249 tracker npm with build

igorkol:


index.js
```diff
@@ -0,0 +1,105 @@
...
+ * returns void, but it runs webpack which builds the colibrius.js file in cwd /dist
+ */
+module.exports = function build(ldSettingsFromDB, ldTrackersFromDB) {
+	const webpack = require('webpack');
+	const path = require('path');
+	const TerserPlugin = require('terser-webpack-plugin');
+	const packageJSON = require('./package.json');
+
+	const webpackSettings = {
+		devtool: process.env.STAGE === 'prod' ? false : 'source-map',
```
Getriax, line 14: I'd suggest going for another function parameter like `webpackOptions` that would have a value like `isProd` instead of relying on environment values. The builder can be run on different environments, that not necessarily have a `STAGE` env variable.

Getriax: Also, what's with the formatting of that JSON. Looks like the prettier isn't working here? It's hard to read it.

index.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * NPM package bundling function
+ * @param {Partial<OwnLayer>} ldSettingsFromDB LittledataLayerBundleSettings received from ld-news
+ * @param {LittledataTrackers} ldTrackersFromDB LittledataEnabledTrackers received from ld-news
+ * returns void, but it runs webpack which builds the colibrius.js file in cwd /dist
+ */
+module.exports = function build(ldSettingsFromDB, ldTrackersFromDB) {
+	const webpack = require('webpack');
```
Getriax, line 8: Let's move the imports to the top of the file, they don't have to be reinitialised with each function run.

index.js
```diff
@@ -0,0 +1,105 @@
+/**
```
Getriax, line 1: Since there is a chance we'll be using this builder with typescript, it would be nice to either make it `index.ts` with (requires tsc before building package) or at least to define `index.d.ts` that will export types definitions for that file.

Getriax (changes requested):
I've changed the base to the branch it was derived from.

After addressing the review changes. We also need to change the circleci pipeline (introduce npm package build and publish workflow) - if you need help with that let me know.

TheGame2500 (changes requested):
changes look ok, but you can't delete the dist folder, there are still old shops on manual depending on it.

Getriax (approved):
LGTM 🚀

igorkol:
@fstefan7 yes, but not dist files - just dist folder. files ended up in it only because i have tested something and then i erased them in some of the next pushes in the PR to clean that up

index.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+export default function build(
+	ldSettingsFromDB: Partial<OwnLayer>,
+	ldTrackersFromDB: LittledataTrackers,
+	buildConfig: BuildConfig,
+): Promise<BuildResult> {
+	const outputDir = path.resolve(buildConfig.rootDir || __dirname, 'bundle-dist', buildConfig.subId);
+
+	const webpackSettings: webpack.Configuration = {
+		devtool: buildConfig.environment === 'development' ? 'source-map' : false,
```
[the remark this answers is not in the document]
igorkol: resolved, thx

PR #385, opened 2022-08-02
re-enabled individual colibrius trackers (g,s,f)

igorkol:


TheGame2500 (approved):

src/segmentTracker/colibrius-s.ts
```diff
@@ -11,12 +17,14 @@ import { retrieveAndStoreFacebookIds } from '../common/helpers/retrieveAndStoreF
 import { urlToQueryStringObject } from '../common/helpers/urlToQueryStringObject';
 import { retrieveAndStoreUTMs } from '../common/helpers/retrieveAndStoreUTMs';
 import { adblockDetect } from '../common/adblockDetect';
+import { buildingIndividualTracker } from '../../types';
```
Getriax, line 20: Why is it imported from `types.ts`? It's a global variable like `LittledataLayerBundleSettings` which should be just defined as global in `index.d.ts`

types.ts
```diff
@@ -197,3 +197,6 @@ export interface BuildResult {
 	directory: string;
 	filename: string;
 }
+
+type BuildingIndividualTracker = boolean;
+export const buildingIndividualTracker: BuildingIndividualTracker = false;
```
Getriax, line 202: You can remove it, after defining a global in `index.d.ts`

src/facebookTracker/colibrius-f.ts
```diff
@@ -26,3 +28,5 @@ export const loadFacebookTracker = (): void => {
 		});
 	});
 };
+
+if (buildingIndividualTracker === true) loadFacebookTracker();
```
Getriax, line 32: Maybe change the name of that to suggest that it's a boolean and just use `if (isBuildingIndividualTracker) load...`

src/facebookTracker/colibrius-f.ts
```diff
@@ -1,15 +1,17 @@
...
 import { sendFbEvent, trackEvents } from './helpers/trackEvents';
 import { LittledataPixel } from '../littledataPixel/LittledataPixel';
+import { buildingIndividualTracker } from '../../types';
 
 declare let window: CustomWindow;
 
-export const loadFacebookTracker = (): void => {
-	getTrackingConsent('fbp').then(() => {
+export const loadFacebookTracker = async (): Promise<void> => {
+	await ensureLittedataLayerWithBundledSettings();
```
Getriax, line 13: Would suggest moving this to the if individual section before the load function is invoked. Because running this one is not necessary for each of the trackers for bundled version. It's already called there once, before bootstraping.

src/segmentTracker/colibrius-s.ts
igorkol: because it creates issues with undefined value of buildingIndividualTracker in the bootstrap file test. when it is left as is, there are no issues

PR #392, opened 2022-08-17
DPD-251 | s3 upload

Getriax:


TheGame2500 (approved):

test/lib/bundleManager.test.ts
```diff
@@ -0,0 +1,74 @@
...
+chai.use(require('sinon-chai'));
+
+const { expect } = chai;
+
+describe('ShopifyScriptBundle', () => {
+	const awsS3ClientMock = {
+		send: sinon.spy(),
+	};
+
+	it('Should generate bundle and upload the files to s3', async () => {
```
igorkol, line 15: maybe have a few smaller tests instead of one big one? like one for generating bundle, one for creating and confirming the path, one for the upload...
that way, it would be easier to debug code and manage the tests in future

index.ts
```diff
@@ -2,24 +2,25 @@ import webpack from 'webpack';
...
+import { BundleManager } from './lib/bundleManager';
 
 /**
  * NPM package bundling function
  * @param {Partial<OwnLayer>} ldSettingsFromDB LittledataLayerBundleSettings received from ld-news
  * @param {LittledataTrackers} ldTrackersFromDB LittledataEnabledTrackers received from ld-news
- * @param {BuildConfig} buildConfig options object for build settings
+ * @param {BundleConfig} bundleConfig options object for bundle settings
  * runs webpack which builds the colibrius.js file in cwd {buildConfig.rootDir || __dirname/bundle-dist/{buildConfig.subId}
  * returns {Promise<BuildResult>} - { directory: directory where the scripts are present, filename: main file name }
```
igorkol, line 14: it would be good if you could update the docs in the files that you changed (for ex. what build fn returns)

.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ slack-notifications: &slack-notifications
 jobs:
   install_npm:
     docker:
-      - image: circleci/node:12
+      - image: circleci/node:14
```
igorkol, line 16: why did you have to bump the node version? what is the dependency?

igorkol (changes requested):
generally, a very good code and a great change. only a few comments left

.circleci/config.yml
Getriax: It's an issue with `fs.promises.rm` which became available only with node v14. I could've used different approach to rm but since we use 14 already across all other server side services, this bump is not an issue.

test/lib/bundleManager.test.ts
Getriax: We can do that, but this PR is not about adding tests, I only tested the stuff I added here which is s3 upload. The test has a few lines of code, but it's required here to check if multiple files are uploaded correctly.

igorkol (approved):

PR #398, opened 2022-08-30
DPD-248 resolving scroll/swipe issue with repeated list views and product clicks on collections page

igorkol:


TheGame2500 (changes requested):
As discussed, let's use a dictionary

src/common/productListViews.ts
```diff
@@ -24,36 +26,38 @@ const millisecondsForProductAPI = window?.LittledataLayer?.productDelay ?? 2500;
...
-			if (isProductViewed && isTrackerOriginSame) {
-				return;
-			}
+			if (isProductInLog) return;
+
 			if (isProductVisible) {
 				// we are checking if product impressions array exists
 				// to prevent duplicates when trackers are bundled
-				if (!isProductViewed)
+				if (!isProductViewed) {
```
TheGame2500, line 50: @igorkol how about `if (isProductVisible & !wasProductViewed)` at line 47? no need to nest and `wasProductViewed` is a bit clearer terminology IMHO

TheGame2500 (approved):

PR #436, opened 2023-01-09
DPD-428 Injected gtag script loading, removed dependancy on ga

igorkol:
ty page script and tests still wip


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "15.2.5",
+  "version": "16.0.0",
```
lageorgem, line 2: We should do bumps on the staging branch, before deploying

src/common/helpers.ts
```diff
@@ -124,9 +124,26 @@ export const advertiseLD = (app: string): void => {
 	}
 };
 
-export function retrieveAndStoreClientId(): void {
-	waitForGaToLoad();
-}
+export const retrieveGtagIds = async (idType: GtagIdType): Promise<string | undefined> => {
+	return new Promise((resolve) => {
+		window.gtag('get', getGAId(), idType, resolve);
+	}).then((val: string) => val);
```
lageorgem, line 130: `.then((val: string) => val)` I think this part is redundant

src/common/helpers.ts
```diff
@@ -124,9 +124,26 @@ export const advertiseLD = (app: string): void => {
...
+};
+
+export const storeGaIdsToTM = async (): Promise<void> => {
```
lageorgem, line 133: TM as in transaction monitor? Technically i think we store this to local storage

src/common/helpers.ts
```diff
@@ -146,39 +163,6 @@ export const documentReady = (callback: CallableFunction): void => {
...
-			return;
-		}
-
-		if (callback) return callback();
-		setCustomTask(trackers[0]);
-
-		return setClientID(getGAClientId(trackers[0]), 'google');
-	};
-
-	waitForObjectToLoad(hasGaTrackers, foundCallback);
```
lageorgem, line 179: Are there any more uses for `waitForObjectToLoad` now?

src/gaThankYouPageTracker/gaThankYouPageTracker.ts
```diff
@@ -34,47 +34,45 @@ export const loadGaThankYouTracker = (): void => {
...
 		sendAdsTransactionConversion(googleAdsConversionIds, total, currency);
 };
 
+const addClientIdToLittledatalayer = async () => {
+	const idResult = await retrieveGtagIds('client_id');
+
+	window.LittledataLayer = {
+		...window.LittledataLayer,
+		attributes: {
+			'google-clientID': idResult,
```
lageorgem, line 61: What about the other note attributes? Isn't this just gonna overwrite them?

src/gaTracker/colibrius-g.ts
```diff
@@ -16,17 +10,14 @@ declare let window: CustomWindow;
...
-				window.LittledataLayer.isDawnThemeActive = window?.Shopify?.theme?.name === 'Dawn';
-				sendPageview();
-				trackEvents();
-			});
+		await storeGaIdsToTM();
+		pageView(() => {
+			window.LittledataLayer.isDawnThemeActive = window?.Shopify?.theme?.name === 'Dawn';
+			sendPageview();
+			trackEvents();
 		});
```
lageorgem, line 21:
```suggestion
	await getTrackingConsent('google');
	initGtag();
	advertiseLD('Google Analytics');
	await storeGaIdsToTM();
	pageView(() => {
		window.LittledataLayer.isDawnThemeActive = window?.Shopify?.theme?.name === 'Dawn';
		sendPageview();
		trackEvents();
	});
```

I suppose at this point you might as well do this

src/gaTracker/helpers.ts
```diff
@@ -15,57 +15,66 @@ import { event_category } from './constants';
 
 declare let window: CustomWindow;
 
-const getWebPropertyId = () => window.LittledataLayer.webPropertyId || window.LittledataLayer.webPropertyID; // spelling prior to LittledataLayer v9.0.1
-const getGA4Id = () => window.LittledataLayer.measurementId;
+const getWebPropertyId = (): string => window.LittledataLayer.webPropertyId || window.LittledataLayer.webPropertyID; // spelling prior to LittledataLayer v9.0.1
+
+export const getGA4Id = (): string => window.LittledataLayer.measurementId;
+
+export const getGAId = (): string => getGA4Id() ?? getWebPropertyId();
```
lageorgem, line 22: the nullish coalescing operator only allows the left-hand side when it's not null or undefined. This means we can allow values such as `0` and `''`, which are falsy but not nullish. Is that intended?

src/segmentTracker/colibrius-s.ts
```diff
@@ -23,11 +23,11 @@ declare let window: CustomWindow;
 
 export const loadSegmentTracker = async (): Promise<void> => {
 	LittledataPixel.init({ pixelId: window.LittledataLayer.writeKey });
-	await getTrackingConsent('segment').then(() => {
+	await getTrackingConsent('segment').then(async () => {
```
lageorgem, line 26: Same story here as colibrius-g, you can get rid of the `.then` and reduce the nesting level by 1

src/segmentTracker/colibrius-s.ts
```diff
@@ -23,11 +23,11 @@ declare let window: CustomWindow;
...
 		initSegment();
 		advertiseLD('Segment');
 		identifyCustomer();
-		retrieveAndStoreClientId();
+		await storeGaIdsToTM();
 		retrieveAndStoreUTMs();
 
 		documentReady(() => {
```
lageorgem, line 33: Speaking of nesting levels, this is turning into a real callback hell. Could you move the `pageView` and `window.analytics.ready` callbacks into their own functions and reference that? 
Bonus points for jsdoc.

lageorgem (commented):
Some minor comments. JSDocs on functions you worked on would be appreciated

package.json
igorkol: we can, sure, but then we will always have a deployment error in the PRs because we didn't bump the version in the PR. wdyt @lageorgem ? maybe we could do the version bump in the PRs and use something else instead of release-it for the master deployment? because we only use release-it for the notes and it anyway creates an issue with the (patch) version bumping as it would merge many minor changes in one small (patch) version change...

src/common/helpers.ts
igorkol: i thought so as well, but i don't get the value as the output unless i have `then` to return it

src/common/helpers.ts
igorkol: it is a two step process in the `setClientID`: first we store the values in the LittledataLayer and then we push to TM. since TM is the whole purpose (final destination) of saving the id, to me it made sense to name the method like this. if this `storeGaIdsToTM` is not run, nothing will be sent to TM. do you have another idea of how to name it @lageorgem that would be clearer on the purpose of it?

src/common/helpers.ts
igorkol: yes: waiting for FB ids, Braze ids, and some other stuff. it would be too much of refactoring if I would remove the fn now

src/gaTracker/colibrius-g.ts
igorkol: good point. done

src/gaTracker/helpers.ts
igorkol: good point. however, my idea was just to return ga4 property id (which comes from ld-news and LDlayer anyway) if it is not undefined or null. i don't expect `0` but you are right thinking that by mistake, someone might send `''` instead of the actual value from layer. that might happen only if the entry validation in the onboarding fails as well as probably a couple of other steps in the ld-news before the entry is saved to mongo. so, do we want to make it more complicated here with the condition?

src/segmentTracker/colibrius-s.ts
igorkol: yep

src/segmentTracker/colibrius-s.ts
igorkol: i wouldn't do that now because analytics.ready part i think we will have to deal with differently soon enough because of the new version of the segment script and the old one being depleted. it's just much refactoring for both of those you mentioned and a scope creep for this card

src/common/helpers.ts
lageorgem: since the actual storing is not done in this function, it makes no sense to name it as such

src/gaThankYouPageTracker/gaThankYouPageTracker.ts
lageorgem: my point is, if `window.LittledataLayer` looks like this:

```typescript
window.LittledataLayer = {
   attribute1: 'asd',
   somethingElse: true,
   attributes: {
     'fbc-clientID': 'fb.1.12345.567889'
   }
}
```

then your operation would cause it to look like this.

```typescript
window.LittledataLayer = {
   attribute1: 'asd',
   somethingElse: true,
   attributes: {
     'google-clientID': '1234567.3456789'
   }
}
```

You're not destructuring existing properties from `LittledataLayer.attributes`, but instead overwriting it. Is it intended behaviour?

src/segmentTracker/colibrius-s.ts
lageorgem: 5 minute job. See my commit.

src/gaTracker/helpers.ts
lageorgem: Defensive programming is better than debugging. Not complicating anything since you're just replacing `??` with `||`

src/common/helpers.ts
lageorgem: That sounds unlikely. all this does is take the resolved value from the promise and return it again.

package.json
lageorgem: Well, bundling isn't really used anyway right now. Yeah sure, if you wanted to test the bundle version of this PR, you would bump it here and also make a PR to bump it on ld-news, so they all work together on staging. Since that doesn't happen, I see no reason to change the release procedure yet

src/gaThankYouPageTracker/gaThankYouPageTracker.ts
igorkol: Yes, as Ldlayer doesn't exist on the typ. We are building it from the query params. It is not the same layer pertained from the frontend before atc. Nothing will be overwritten here

igorkol:
@lageorgem codeclimate is failing because of the refactor you did in the segment tracker. would you like to resolve code complexity and write additional tests now, or should we leave it for some next iteration of tech debt cleanup?

lageorgem:
> @lageorgem codeclimate is failing because of the refactor you did in the segment tracker. would you like to resolve code complexity and write additional tests now, or should we leave it for some next iteration of tech debt cleanup?

@igorkol The cognitive complexity was there before, I just marked as wontfix. Tests will have to be written

lageorgem (approved):
Wonderful

PR #438, opened 2023-01-13
feature/DPD-470_disable_benchmarks

MihaiLittledata:


lageorgem (approved):
Looks good, @igorkol pls also take a look.

igorkol:
@MihaiLittledata you missed commenting out `LittledataPixel` in the main files, which means that it is still initiated and sends pageviews. So, `LittledataPixel.init` should be additionally removed from the main three colibrius scripts and the two thankyou page scripts

PR #442, opened 2023-01-26
DPD-475 Adding underscore to note_attribute IDs

igorkol:


src/segmentTracker/helpers/addEmailToEvents.ts
```diff
@@ -19,7 +19,7 @@ export const addEmailToEvents = ({ payload, next }: MiddlewareOptions): void =>
 };
 
 export const addEmailToProperties = (properties: LooseObject): LooseObject => {
-	const email = window.analytics?.user?.().traits().email || window.LittledataLayer.attributes?.['email-clientID'];
+	const email = window.analytics?.user?.().traits().email || window.LittledataLayer.attributes?.['_email-clientID'];
```
lageorgem, line 22: what about instances where the attribute is already the legacy one? Shouldn't we look for both?

src/segmentTracker/helpers/addUTMToEvents.ts
```diff
@@ -26,7 +26,7 @@ export const addUTMToEvents = ({ payload, next }: MiddlewareOptions): void => {
  * @returns Properties payload with the UTMs
  */
 export const addUTMToProperties = (properties: LooseObject): LooseObject => {
-	const utmString = window.LittledataLayer?.attributes?.['UTM-clientID'];
+	const utmString = window.LittledataLayer?.attributes?.['_UTM-clientID'];
```
lageorgem, line 29: what about instances where the attribute is already the legacy one? Shouldn't we look for both?

src/segmentTracker/helpers/addEmailToEvents.ts
igorkol: yes, i missed it during search

src/segmentTracker/helpers/addUTMToEvents.ts
igorkol: yes, i missed it during search

Getriax (approved):
Looks ok, though there is a lot of string declarations for cids, maybe consider moving them to enums? 

```typescript
enum clientID {
  google = '_google-clientID',
  segment = '_segment-clientID',
  ...etc
}
```

Also maybe it would be worth declaring one getter for attributes? Like method that would accept new cid name and return it looking for legacy as well?

```typescript
export const getLittledataAttribute(cidName: clientID): string | undefined {
   return window.LittledataLayer?.attributes?.[cidName] || window.LittledataLayer?.attributes?.[cidName.substring(1)];
}
```

igorkol:
good point about the method for legacy stuff @Getriax . however, that block of code that sets the clientIDs is already spaghetti and i would like to make as minor changes to it as possible until it gets split into smaller chunks and refactored

as for the enums, i did make those where possible in another PR (https://github.com/littledata/shopify-tracker/pull/447) while these clientID-related ones i don't think would benefit from an enum as there are strict types around these

PR #451, opened 2023-02-24
DPD-499 Measuring impact of gtag clientID resetting in GA

igorkol:


src/common/helpers.ts
```diff
@@ -145,6 +146,11 @@ export const fetchAndStoreGtagIds = async (): Promise<void> => {
 
 	if (typeof idResult === 'string') {
 		setClientID(idResult, 'google');
+		await LittledataPixel.sendEvent({
```
TheGame2500, line 149: @igorkol I see fetchAndStoreGtagIds is used in collibrius-s as well. it will throw an error because LittledataPixel is not initialised. Can you wrap this in a try / catch to just swallow the error. i don't want to report from segment setups so it's fine i reckon

TheGame2500 (changes requested):

src/common/helpers.ts
igorkol: right! on it

PR #467, opened 2023-03-16
DPD-522 Removed currency, customer and page type dependencies on liquid

igorkol:


src/common/helpers/getCustomerFromHTML.ts
```diff
@@ -0,0 +1,25 @@
+import { CustomWindow } from '../../../index.d';
+import { IScrapedCustomer } from '../../../types';
+import { debug } from 'webpack';
+
+declare let window: CustomWindow;
+
+export const getCustomerFromHTML = (): IScrapedCustomer | undefined => {
+	const webPixelScript = window.document.getElementById('web-pixels-manager-setup')?.innerHTML;
```
lageorgem, line 8: Add a big try/catch on this, as it feels flaky and it may fail. it can return on catch

lageorgem (approved):

PR #471, opened 2023-03-31
DPD-522 Remove dependancy on liquid

igorkol:


igorkol:
this is an existing code on Staging. i reinstated a merged PR and made a patch which got forgotten in the previous PR for which i already wrote tests. i didn't plan on writing additional tests only to cover this diff

lageorgem (approved):

PR #472, opened 2023-04-03
DPD-510 | Disable selected client side events by destination setting

Getriax:


lageorgem (commented):
Looks great, but asking for a look from @igorkol

lageorgem (approved):

PR #488, opened 2023-05-04
DPD-547 Cleaning up and tests

igorkol:


igorkol:
@TheGame2500 please check that I haven't messed something up for the legacy users

PR #489, opened 2023-05-08
bugfix/DPD-535_Order_ID

lageorgem:


TheGame2500 (approved):

igorkol (changes requested):
changes are needed in src/common/buildGTMParamsForTYPage.ts as well.

also, no tests to cover this?

PR #491, opened 2023-05-09
DPD-609 Added myshopify.com store name to GA4 events

igorkol:


src/gaTracker/helpers.ts
```diff
@@ -80,6 +80,7 @@ export const ga4EventParams = (): GA4EventParameters => {
 		// littledata_client_id: window.LittledataLayer?.attributes?.['_google-clientID'],
 		sent_from: 'Littledata script',
 		send_to: window.LittledataLayer.measurementId,
+		store_name: window.Shopify?.shop,
```
lageorgem, line 83:
```suggestion
		...(window.Shopify?.shop && { store_name: window.Shopify?.shop }),
```

what if it's undefined?

lageorgem (approved):

PR #498, opened 2023-05-22
feature/rebuild

lageorgem:


src/rebuild/destination-service/strategies/GA4DestinationStrategy.ts
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	public async getClientID() {
+		const clientIdPromise = new Promise((resolve) => {
+			window.gtag('get', this.settings.measurementId, 'client_id', resolve);
+		});
+		const sessionIdPromise = new Promise((resolve) => {
+			window.gtag('get', this.settings.measurementId, 'session_id', resolve);
+		});
+		const sessionCountPromise = new Promise((resolve) => {
```
igorkol, line 64: @lageorgem where did you get this from? ga docs don't mention this is possible. afaik, the only way is to get it from the cookie

lageorgem: Left as a placeholder, didn't really expect it to work. As per docs, only accepted values are client_id, session_id and gclid.

PR #507, opened 2023-06-08
DPD-637 Order payload contained in the GA4 TY page event

igorkol:


src/gaTracker/helpers.ts
```diff
@@ -472,4 +474,34 @@ function convertProductsToGa4Format(products: IDetail[], sendIndex: boolean): GA
...
+			item_id: product.variant_id.toString(),
+			item_name: product.title,
+			item_brand: product.vendor,
+			item_variant: product.variant_title,
+			item_sku: product.product_id.toString(),
+			item_variant_id: product.variant_id.toString(),
+			price: +product.price,
+			discount: +product.applied_discounts?.[0]?.amount || 0,
+			quantity: product.quantity,
+			// TODO: add index key which should be product.list_position consistent from the catalog page
```
lageorgem, line 492: what about this todo? i think we should have this saved somewhere in localstorage

src/gaTracker/helpers.ts
```diff
@@ -416,15 +417,16 @@ function sendSelectContentEvent(product: IDetail, self: TimeBombHTMLAnchor, open
...
-	const shopifyCheckoutObj = window.Shopify?.checkout;
+export const sendThankYouEvent = (): void => {
+	const shopifyCheckoutObj: IShopifyOrderConfirmation = window.Shopify?.checkout;
 	const total = shopifyCheckoutObj?.total_price;
 	// const currency = shopifyCheckoutObj?.currency;
 	const value = parseInt(total, 10);
 
 	if (hasGA4()) {
-		GoogleAnalyticsEventHandler.instance.sendEvent('Checkout', {
+		GoogleAnalyticsEventHandler.instance.sendEvent('Thank You page', {
```
lageorgem, line 427: Is the 'Checkout' event obsolete? Can we just rename it without consequences?

src/gaTracker/helpers.ts
igorkol: @lageorgem yes, we do, but there is no point in sending this parameter with the event which is not an e-commerce event (like `purchase` is) and also we are not sending the position with the server events.
so, it is intentionally left as a reminder that first the server events have to have this and also that potentially this event above should be renamed to a "purchase" event (based on discussion on slack)

src/gaTracker/helpers.ts
igorkol: @lageorgem The `checkout` event doesn't exist within the ga4 standard nomenclature. when I added this event initially, it was just a copy of a GA3 event (because in GA3 this is the last event after the previous 3 checkout steps, within the same event action called "checkout").
however, now @LittleLeonard has insisted for some time already for me to rename this event to the "thank you page". also, @edwardupton said this as well. I did warn previously in DPD Slack convo that this naming change might break tracking for some clients that are pulling in this GA4 event. My opinion is that this should be announced to the clients and then changed after some time given for adaptation

lageorgem (approved):

PR #508, opened 2023-06-12
feature/DPD-618-pinterest-thank-you-page

SMihail:
#### Changelog:
* added pinterest tracking on the thank you page

#### Checklist:
- [ ] Code review was done by the following engineers:
- [ ] PR was validated to be ready to merge by the following QA engineer:

[DPD-618]: https://littledata.atlassian.net/browse/DPD-618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/eventHandler/baseEventHandler.ts
```diff
@@ -51,4 +46,8 @@ export abstract class BaseEventHandler<Body = any, Options = any, PageViewData =
 	isEventDisabled(name: string): boolean {
 		return (this.disabledEvents || []).map((ev) => ev.toLowerCase()).includes(name.toLowerCase());
 	}
+
+	protected abstract handleEvent(name: string, body: Body, options?: Options): Promise<void>;
```
lageorgem, line 50: Why move these declarations here? Abstract methods should be at the top before any implementation

lageorgem: bump?

SMihail: webstorm automatically moves the two methods when using 
image

Disabling to keep the order of the methods as requested

lageorgem (approved):

PR #551, opened 2023-07-26
DPD-518 | Handle ga4 login event

Getriax:
#### JIRA Issue: [](url)
https://littledata.atlassian.net/browse/DPD-518

#### Changelog:
* Send ga4 login event on login page view

#### Checklist:
- [x] Code review was done by the following engineers:
- [x] PR was validated to be ready to merge by the following QA engineer:


src/gaTracker/events/LoginEvent.ts
```diff
@@ -0,0 +1,18 @@
...
+			return;
+		}
+
+		await GoogleAnalyticsEventHandler.instance.sendEvent('login', {
+			...ga4EventParams(),
+		});
+	}
+
+	private shouldTrigger() {
+		return hasGA4() && window.location.href.includes('account/login');
```
lageorgem, line 16: this only tracks the fact that the user is on the login page, not actual login. I suggest adapting [this piece of code](https://github.com/littledata/shopify-tracker/pull/498/files#diff-f1c36c550e39d009cd81542602c714582c1b54b3648c7ad2709fe46279cf7251R13) to store a state in localStorage. when customer is authenticated but localStorage does not contain that state, you can assume the user has authenticated in the meantime

src/gaTracker/events/LoginEvent.ts
```diff
@@ -0,0 +1,47 @@
...
+		if (window.location.href.includes('account/login')) {
+			// clear this item on login page
+			window.sessionStorage.removeItem('ld_authenticated');
+
+			return false;
+		}
+
+		const isAlreadyAuthenticated = window.sessionStorage.getItem('ld_authenticated');
+
+		if (isAlreadyAuthenticated && isAlreadyAuthenticated === 'true') {
```
lageorgem, line 32: `Boolean(isAlreadyAuthenticated)`

lageorgem (approved):

src/gaTracker/events/LoginEvent.ts
```diff
@@ -0,0 +1,47 @@
...
+			return;
+		}
+
+		await GoogleAnalyticsEventHandler.instance.sendEvent('login', {
+			...ga4EventParams(),
+		});
+	}
+
+	private shouldTrigger() {
+		if (!hasGA4() || !window.location.href.includes('account')) {
```
lageorgem, line 19: This makes the assumption that after login, you will be redirected to the `/account` page, and only here you will trigger the login event. This may not always be the case, as it depends on the return_url set on the login page:

https://george-littledata.myshopify.com/account/login?return_url=/products/tricou-steaua

I think you should revise the logic to something like

`if (!isAlreadyAuthenticated && customerId)`

which should tell us that we never marked this user as logged in, but they are in fact logged in.

True, the caveat will be sending a bunch of login events for users that are already logged in, but this should happen only after deployment or installation

lageorgem (changes requested):

lageorgem (approved):

lageorgem:
@Getriax conflicts

lageorgem (approved):

PR #553, opened 2023-08-01
DPD-519/ga4-signup-event

SMihail:
#### Changelog:
* track signup

#### Checklist:
- [x] Code review was done by the following engineers:
- [x] PR was validated to be ready to merge by the following QA engineer:


src/gaTracker/events/SignupEvent.ts
```diff
@@ -0,0 +1,39 @@
...
+
+declare let window: CustomWindow;
+
+export class SignupEvent {
+	async track() {
+		if (!this.shouldTrigger()) {
+			return;
+		}
+
+		await GoogleAnalyticsEventHandler.instance.sendEvent('register', {
```
lageorgem, line 13: event name should be [sign_up](https://developers.google.com/analytics/devguides/collection/protocol/ga4/reference/events#sign_up)

src/gaTracker/events/SignupEvent.ts
```diff
@@ -0,0 +1,39 @@
...
+			...ga4EventParams(),
+		});
+	}
+
+	private shouldTrigger() {
+		if (!hasGA4() || !window.location.href.includes('account')) {
+			return false;
+		}
+
+		if (window.location.href.includes('account/register')) {
```
lageorgem, line 23: The registration flow is made more complicated by the confirmation email you receive. I think the algorithm should be as follows:

1. If we are on `account/register`, add the `localStorage` `registering`
2. if we are on `account/login`, remove the `localStorage` `registering` - indicating the user will instead opt for regular authentication
3. if the `localStorage` `registering` key exists AND `window.ShopifyAnalytics?.meta?.page?.customerId` is numeric, then it means that: 
    a) the user has authenticated since visiting the registration page
    b) the user hasn't gone through the `account/login` page - so likely they registered and then clicked the confirmation URL.

lageorgem (changes requested):

src/gaTracker/events/SignupEvent.ts
SMihail: https://developers.google.com/analytics/devguides/config/admin/v1/rest/v1alpha/accounts/provisionAccountTicket

`sign_up` or `provisionAccountTicket`? unclear from the admin api docs

src/gaTracker/events/SignupEvent.ts
```diff
@@ -0,0 +1,51 @@
...
+
+	private shouldTrigger(): boolean {
+		if (!hasGA4()) {
+			return false;
+		}
+
+		if (window.location.href.includes('account/login')) {
+			// indicate the user opts for regular authentication
+			// by being present on the login page
+			window.sessionStorage.removeItem('ld_registered');
```
lageorgem, line 26: `sessionStorage` exists per browser tab, and is deleted when closing the browser. It doesn't survive like the auth cookie would. switch to `localStorage`

https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage

src/gaTracker/events/SignupEvent.ts
```diff
@@ -0,0 +1,51 @@
...
+
+			return false;
+		}
+
+		if (window.location.href.includes('account/register')) {
+			window.sessionStorage.removeItem('ld_registered');
```
lageorgem, line 32: i think you should add this storage item on registering, not remove it. take a look at my previous comment again.

src/gaTracker/events/SignupEvent.ts
```diff
@@ -0,0 +1,51 @@
...
+			return false;
+		}
+
+		const isRegistered = window.sessionStorage.getItem('ld_registered');
+
+		// client id appears only after the user is authenticated
+		// if cid appears and isRegistered is true, then the user registered and afterwards clicked the link in the confirmation e-mail, thus skipping login page
+		const cid = window.ShopifyAnalytics?.meta?.page?.customerId;
+
+		if ((isRegistered && isRegistered === 'true') || (isRegistered && typeof cid === 'number')) {
```
lageorgem, line 43: you probably don't need to do the early return like this here. I'd argue it's better to do something like this

```typescript
if (isRegistering && cid) { 
  return true 
}
```

Also pls change all `registered` to `registering`, since it's an ongoing state. The logic would be to mark the fact that the user has begun the registration process, and send the event when the user has actually logged in, with the exception being that you remove the `registering` state on the login page

src/gaTracker/events/SignupEvent.ts
lageorgem: sign_up

src/gaTracker/events/SignupEvent.ts
```diff
@@ -0,0 +1,51 @@
...
+
+	private shouldTrigger(): boolean {
+		if (!hasGA4()) {
+			return false;
+		}
+
+		if (window.location.href.includes('account/register')) {
+			window.localStorage.setItem('ld_registering', 'true');
+
+			return true;
```
lageorgem, line 26: no, event should not be triggered when we are on the register page. we only need to set the flag

src/gaTracker/events/SignupEvent.ts
```diff
@@ -0,0 +1,51 @@
...
+
+		// client id appears only after the user is authenticated
+		// if cid appears and isRegistering is true, then the user clicked the link in the confirmation e-mail after registering, thus skipping login page
+		const cid = window.ShopifyAnalytics?.meta?.page?.customerId;
+
+		if (isRegistering && cid) {
+			return false;
+		}
+
+		window.localStorage.setItem('ld_registering', 'true');
```
lageorgem, line 47: not sure this is needed here

src/gaTracker/events/SignupEvent.ts
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		const isRegistering = window.localStorage.getItem('ld_registering');
+
+		// client id appears only after the user is authenticated
+		// if cid appears and isRegistering is true, then the user clicked the link in the confirmation e-mail after registering, thus skipping login page
+		const cid = window.ShopifyAnalytics?.meta?.page?.customerId;
+
+		if (isRegistering && cid) {
+			return false;
```
lageorgem, line 44: if the user is registering and we are authenticated, we **do not want** to trigger the event? change this to true

src/gaTracker/events/SignupEvent.ts
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		window.localStorage.setItem('ld_registering', 'true');
+
+		return true;
```
lageorgem, line 49: a default action should be do not trigger, so false

lageorgem (changes requested):

lageorgem (approved):

lageorgem (approved):

